Issue #30, opened 2025-04-06
Add evals scaffolding


dcramer:
Some inspo:

https://github.com/cloudflare/agents/blob/main/packages/agents/evals/README.md

https://github.com/neondatabase-labs/mcp-server-neon/tree/shridhar/mcp-sse/src/tools-evaluations

Tentative plan:

- Wire up a basic system prompt
- Connect our MCP to it (likely via stdio)
- Get to building the Eval suite

We can run it manually or automatically depending on cost

Gonna say blocks on #28

PR #32, opened 2025-04-06
Add basic eval scaffolding

dcramer:
Add the following to .env:

- `OPENAI_API_KEY`
- `SENTRY_AUTH_TOKEN`

And then run:

```
pnpm eval:dev
```

Fixes GH-30

### DIFF
--- a/.github/workflows/eval.yml
+++ b/.github/workflows/eval.yml
@@ -1,0 +1,67 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  conditional_check:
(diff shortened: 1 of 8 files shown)


dcramer:
FYI I had to downgrade vitest to 2.x as evalite is broken in 3.x

I will also note that evalite may be a problematic dep. The issue tracker makes it look not super well maintained (theres some pretty basic issues hanging), and its got a number of third party deps that are a little complicated.

A future better solution here probably would be to monorepo this project, and push evalite/etc into its own package so it wont conflict.

dcramer:
This wont yet run correctly in prod til I can setup a bot account on Sentry

PR #38, opened 2025-04-07
feat: React+Tailwind

dcramer:
Swaps our usage to be Vite-based per https://developers.cloudflare.com/workers/vite-plugin/


dcramer:
1) No clue if this is going to work in prod yet. Need to figure out how to test on a new worker.

2) I'm not totally happy w/ the split of code. Might instead do client/shared/worker

package.json
```diff
@@ -2,18 +2,19 @@
...
+    "deploy": "npm exec wrangler deploy",
     "eval": "evalite",
     "eval:ci": "evalite --threshold=100",
     "eval:dev": "evalite watch",
     "format": "biome format --write",
     "lint": "biome lint",
     "lint:fix": "biome lint --fix",
     "inspector": "pnpx @modelcontextprotocol/inspector@latest",
-    "start": "wrangler dev",
+    "start": "vite preview",
```
dcramer: note: rename this to preview, as its not starting a real server

dcramer:
Pushed this into prod and it def works. Will fix tests and merge.

(still need to figure out why CF preview envs are totally non existant)

PR #39, opened 2025-04-09
Expose autofix tools

dcramer:
Adds the `begin_autofix` and `get_autofix_status` tools.

These are not currently streaming (spec doesnt support it), so we're praying that an LLM will be helpful enough to retry or at least inform the user to check back soon.


dcramer:
No streaming response support yet in CF's MCP adapter. I think it should be doable upstream? Maybe?

dcramer:
Not supported in the spec yet

dbworku:
Awesome!! I was just asking Indragie about this yesterday. Thanks @dcramer

dcramer:
@dbworku fwiw it only semi-works - mostly because autofix takes a few minutes to run, the agents tend to get bored waiting that long and try to do their own thing

ill record a video and publish later tonight, but here was a quick example of me cheating the system

https://x.com/zeeg/status/1914801436480954697

the workflow of its quite nice if you could solve for the streams/delay (just not doable yet)

PR #47, opened 2025-04-13
Various tool improvements

dcramer:


dcramer:
Our evals dont run in true agent mode, so in a production use case the LLM would likely have re-queried, but because we want optimal responses its worth fixing this behavior.

PR #56, opened 2025-04-17
Implement monorepo

### LINKED ISSUE #52: Monorepo
Let's split this into a monorepo so we can cleanly enable at least two or three core packages:

- mcp server
- cloudflare worker
- web application (e.g. the home page)

The mcp server should provide a binary which runs the stdio server.

If splitting the web app from the worker is too complex thats fine.

dcramer:
Migrate to a monorepo so we can cleanly release the MCP server without including the Cloudflare and Web components.

There are two core known issues unresolved:

- The `sentry-mcp` bin does not effectively correctly transpile (#64)
- The docker build has issues building (#65)

Fixes #52


dcramer:
There's one remaining issue afaict:

the `bin` file for the stdio interface doesnt manage transpilation effectively

it looks for the literal raw import, which is extensionless

dcramer:
Honestly I may merge this even without this fix as its the same behavior as today. We just cant release it yet.

dcramer:
Docker build is broken and I have no idea how to resolve it, as the lockfile has arch specific pins

f9643 (approved):

f9643 (approved):

PR #61, opened 2025-04-18
feat: Add an eval for list_tags

dcramer:


dcramer:
Was hoping this would flush out a mistake that Copilot made (where it didnt actually pass in the org slug), but what can you do.

PR #131, opened 2025-04-29
Add default Sentry DSN for stdio dist

dcramer:


dcramer:
This will give us vis into stdio distribution errors

PR #137, opened 2025-05-01
Enable HTTP Streamable support from MCP/Agents SDK

codyde:
Adding support for Cloudflare / MCP [Streamable HTTP](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http)

/sse is kept for backwards compatibility 
/mcp should be used when possible, as new clients support it


dcramer:
Need to look into whats up with types here. EIther 1) somethings wrong in Sentry's wrapper, or 2) somethings wrong in Cloudflare's SDK, or 3) somethings wrong in our ability to write code

PR #146, opened 2025-05-01
ci: Tidy up releases, fix stale version.ts issue

BYK:


.github/workflows/merge-jobs.yml
```diff
@@ -51,10 +36,7 @@ jobs:
 
       - name: Package mcp-server
         working-directory: packages/mcp-server
-        run: |
```
dcramer, line 54: wheres build happen?

dcramer:
Wont this still have the same issue where the version is never correct in package.json in main?

.github/workflows/merge-jobs.yml
BYK: Line 29 above. Running build on the root builds all projects in the workspace.

BYK:
>Wont this still have the same issue where the version is never correct in package.json in main?

It will always be the last released version. The value in package.json is managed by Craft releases and the bump-version script which are unchanged. 

We just change where, version.ts file gets it's information which is from the compiler at build time.

PR #154, opened 2025-05-06
feat: Update image and add social preview meta

HazAT:


dcramer:
Wish the logo had the same illustration style and colors :(

dcramer:
got a slightly more color correct image in #252 - just need to update this to get our socials on

Issue #155, opened 2025-05-06
Errors unless providing a specific issue URL


scopendo:
### Environment

Windsurf Version: 1.8.102+next.ac2f9318a9
OS: macOS 15.4.1

### Steps to Reproduce

1. Add sentry-mcp to Windsurf

```json
{
    "mcpServers": {
        "Sentry": {
            "command": "npx",
            "args": ["mcp-remote", "https://mcp.sentry.dev/sse"]
        }
    }
}
```

2. Simple prompt like "what organisations do I have in Sentry?"

Note I tried deleting `~/.mcp-auth` as mentioned on another issue.

### Expected Result

Expected to get a list of the organisations.

### Actual Result

```
MCP Tool: Sentry / list_organizations
Ran with these arguments:
{}
Output
MCP error -32602: Invalid arguments for tool list_organizations: [ { "code": "invalid_type", "expected": "object", "received": "undefined", "path": [], "message": "Required" } ] (Code -32602)

MCP error -32602: Invalid arguments for tool list_organizations: [ { "code": "invalid_type", "expected": "object", "received": "undefined", "path": [], "message": "Required" } ] (Code -32602)
It looks like there was an error when trying to fetch your Sentry organizations. This could be due to a temporary issue or a missing configuration.

To proceed, please ensure:

You are authenticated with Sentry in your environment.
Any required API keys or credentials are set up.
If you have this information or want to try again, let me know! If you want to manually check, you can also visit https://sentry.io/organizations/ to see your organizations directly.
```

### Additional info

Other prompts fail too like:

Prompt: what is the latest issue in Sentry for project [redacted-org-slug]/[redacted-project-slug]

```
I'll check the latest issue in Sentry for the [redacted] mobile app. Let me query Sentry for this information.

MCP Tool: Sentry / list_issues
Ran with these arguments:
{
  "organizationSlug": "redacted-org-slug",
  "projectSlug": "redacted-project-slug",
  "sortBy": "last_seen"
}
Output
Error

There was a 400 error with the your request to the Sentry API.

You do not have access to query across multiple projects. Please select a project for your query.

If you believe this was a genuine error, please report the following to the user for the Sentry team:

Event ID: 109f3fdac020499cb857483d074e766e
```

Note that getting details of an issue by its URL is working fine, which suggests to me that authorisation is fine.  The agent calls "get_issue_details" and is able to give the right summary.

dcramer:
Does windsurf tell you what model it’s using?

[...]

PR #158, opened 2025-05-06
Support regions on API calls

dcramer:
Refs #155

### DIFF
--- a/packages/mcp-cloudflare/package.json
+++ b/packages/mcp-cloudflare/package.json
@@ -4,9 +4,7 @@
   "private": true,
   "type": "module",
   "license": "FSL-1.1-ALv2",
-  "files": [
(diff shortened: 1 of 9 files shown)


dcramer:
followups we should explore:

1. Cache list_organizations in the DO
2. Have all tool calls require an organizationSlug, implying they should always call list_organizations first
3. Make regionUrl required

packages/mcp-cloudflare/package.json
```diff
@@ -4,9 +4,7 @@
   "private": true,
   "type": "module",
   "license": "FSL-1.1-ALv2",
-  "files": [
-    "./dist/*"
-  ],
+  "files": ["./dist/*"],
```
dcramer, line 7: noidea.jpg

PR #164, opened 2025-05-07
Add forced approval flow for MCP intermediary

dcramer:
Address vulnerability where an attacker can effectively bypass authorization once you've already approved a primary client.

https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/265

This also takes the opportunity to bump up the oauth worker bindings, which addresses another security vulnerability (missing verification of redirect_uri).

Note: mcp-remote@latest is required to resolve a bug customers may experience with the oauth change above.


dcramer:
image

dcramer:
Aside, I dont really grok why we are utilizing cookies for this, but I'm basing this on CF's example implementation right now.

It seems to me we should simply look to see if (accessToken, clientId) already exists, and if so that means its already been approved and this should be bypassable.

dcramer:
COOKIE_SECRET deployed to prod

PR #165, opened 2025-05-07
feat: Upgrade JS SDK to `9.16.1` and add durable objects instrumentation

AbhiPrasad:
Durable objects instrumentation support was added in [9.16.0](https://github.com/getsentry/sentry-javascript/releases/tag/9.16.0) of the JS SDK. This PR upgrades `sentry-mcp` to get instrumented with durable objects.

Docs PR for durable objects is here: https://github.com/getsentry/sentry-docs/pull/13626.

In `packages/mcp-cloudflare/src/server/lib/mcp-transport.ts` I had to comment out the `wrapMcpServerWithSentry` usage, because this breaks locally with miniflare. We're following up with the cloudflare team about this (more details here in linear: https://linear.app/getsentry/issue/JS-398/9150-breaks-cloudflare-sdk-usage-in-sentry-mcp#comment-7ce26edb). When we fix this, we'll come back and re-enable it.

`9.16.0` also added support for logs in the Cloudflare SDK, so I enabled that, and added the `consoleLoggingIntegration` (integration is optional while logs is experimental in the SDK) to flush logs from `console.X` calls to Sentry.


packages/mcp-cloudflare/src/server/lib/mcp-transport.ts
```diff
@@ -25,8 +34,27 @@ export default class SentryMCP extends McpAgent<Env, unknown, WorkerProps> {
         userId: this.props.id,
       },
       onToolComplete: () => {
```
dcramer, line 36: do we actually n eed this flush at this point? still no idea why it didnt do anything before

packages/mcp-cloudflare/src/server/lib/mcp-transport.ts
```diff
@@ -25,8 +34,27 @@ export default class SentryMCP extends McpAgent<Env, unknown, WorkerProps> {
...
+    initialScope: {
+      tags: {
+        durable_object: true,
+        mcp_server_version: LIB_VERSION,
+      },
+    },
+    _experiments: {
+      enableLogs: true,
+    },
+    integrations: [Sentry.consoleLoggingIntegration()],
```
dcramer, line 57: does this make console.error go to sentry as a new issue? if so prob gonna remove it as i prefer the more explicit abstraction

packages/mcp-server/src/logging.ts
```diff
@@ -1,4 +1,4 @@
-import { captureException, captureMessage, withScope } from "@sentry/core";
+import { captureException, captureMessage, withScope } from "@sentry/node";
```
dcramer: mcp-server is used on multi-env which is why im not binding this to node

dcramer:
Gonna ship and address comments async.

I reverted the `@sentry/node` changes to the shared package (except in stdio where its always node). It just doesnt feel right to me to suggest node when we have framework bindings, so prefer core.

packages/mcp-cloudflare/src/server/lib/mcp-transport.ts
AbhiPrasad: > does this make console.error go to sentry as a new issue

Right now it only records it as a `error` level log.

Once the SDK is not experimental, we'll merge `consoleLoggingIntegration` with the regular console instrumentation (which is enabled by default), which give users the option to create issues from console errors. For now it will stay separate, and only create logs.

packages/mcp-cloudflare/src/server/lib/mcp-transport.ts
AbhiPrasad: Shouldn't need this flush now that we have durable objects instrumentation. We can keep it around for now, let me test quickly to make sure everything is okay - I'll open another PR in a sec.

packages/mcp-cloudflare/src/server/lib/mcp-transport.ts
dcramer: ah thats what i want (I dont ever want console to create issues, generally speaking)

maybe sometimes id want things showing up as logs to be issues, but by default logger integrations are noisey

AbhiPrasad: > maybe sometimes id want things showing up as logs to be issues, but by default logger integrations are noisey

yeah we're debating between just a boolean option to forward everything, or some callback that allows you to create issues on log message. We'll have this implemented by GA.

dcramer: id just have the server handle it - no reason we cant have a thing that says "turn logs that look like X into issues"

PR #172, opened 2025-05-07
Support event ID lookups on get_issue_details

### LINKED ISSUE #160: Add tool to get specific error event information
Right now there is a `get_issue_details` tool, which obtains the latest error event:
https://github.com/getsentry/sentry-mcp/blob/da9de41fc6b7bac3b4df3543fcbfe0855ed25462/packages/mcp-server/src/tools.ts#L296-L299

There are some cases where we have a specific event id and should be able to get information about that event. Currently that capability does not exist in the MCP server.

dcramer:
Fixes #160


dcramer:
Trying to avoid tool bloat here, so sticking it under the get_issue_details existing tool call. My hope is that with the addition of examples (assuming they dont get truncated), most agents will use the call correctly.

PR #182, opened 2025-05-08
Update names to Seer

ItalyPaleAle:
As the name Autofix is being retired, updating the MCP server to use "Seer" and/or its capability "issue fix" as appropriate


dcramer:
Prob will tweak some of this later as "Sentry's Seer" Isnt really how we think about the persona (e.g. we dont say "Apple's Siri").

ItalyPaleAle:
> Prob will tweak some of this later as "Sentry's Seer" Isnt really how we think about the persona (e.g. we dont say "Apple's Siri").

Agreed. I assumed I needed to put "Sentry" in the sentence to make sure the command to the LLM was clear. However, I have never written code like this before, so I may be off.

Would something like `Use Seer in Sentry and help me analyze and propose a solution for ISSUE_URL.` work better, assuming including Sentry is required indeed?

dcramer:
Yeah probably but I'll dig a bit more. I think given the evals still pass we're in a good sopt, and i think generally "Fix an issue using Seer" or "fix an issue using Sentry" we really want both to work

PR #195, opened 2025-05-15
feat: Upgrade JS SDK to 9.19.0

AbhiPrasad:
resolves https://github.com/getsentry/sentry-javascript/issues/16182


dcramer:
Will merge this tomorrow. I had to revert a dep bump that broke sentry and process.env. All I know is agents broke the latter but didn’t confirm if sentry was the cause of its own failure or not, so need to monitor when I upgrade to ensure data is still coming in.

PR #219, opened 2025-05-27
Add contexts to event responses

### LINKED ISSUE #181: Feature Request: Add some `context` and (some of?) `contexts` fields to issue event response
There's some info I'm looking to grab from individual issue events:
- `contexts` which contains stuff like the user's browser, trace information.
    - Specifically, the trace information is what's most useful when trying to correlate a sentry issue with other parts of our observability stack (in other tools).
- `context` which contains arbitrary user-attached metadata.
    - This has a bunch of extra error context for our errors that would be useful for the LLM.

I'm sure there are more fields outside of these we are omitting for now. This is also at odds with wanting to keep token usage lower where possible. Thoughts @dcramer? Should there perhaps be a flag for "more details" or something so it's optional? 🤔

dcramer:
Fixes #181

### DIFF
--- a/packages/mcp-server/src/api-client/schema.ts
+++ b/packages/mcp-server/src/api-client/schema.ts
@@ -182,6 +182,22 @@ const BaseEventSchema = z.object({
       }),
     ]),
   ),
+  contexts: z
+    .record(
+      z.string(),
+      z
+        .object({
+          type: z.union([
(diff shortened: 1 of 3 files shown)


dcramer:
this may end up noisy and will require state of the art models to make sense of, but thats fine

PR #221, opened 2025-05-27
Refactor tools for simplicity

dcramer:
- list_* to find_*
- search_* to find_*
- organizationSlug is now explicit everywhere
- improved various tool metadata
- add `query` param to find_releases


dcramer:
Primary motivation here is to reduce tool errors. search vs list doesnt add anything, and the implicit organizationSlug either likely was unused or was used regularly incorrectly by various agents.

PR #287, opened 2025-06-21
Consolidate rules into new documentation

dcramer:
This combines rules into a new docs/ folder, and attempts to keep cursor.mdc and Claude.md in sync.


dcramer:
This may have some issues yet but we'll iterate on it.

PR #342, opened 2025-06-30
feat: add overwatch bug prediction gh workflow

suejung-sentry:
Add overwatch bug predictor github workflow to this repo to enable the alpha feature.

Note that this feature is currently being merged with the AI PR review feature (work tracked [here](https://linear.app/getsentry/project/25q2-error-prediction-becomes-a-ai-pr-review-feature-60c79fb6bcf8/issues)). 
Once that project is completed, this github action workflow can be removed (this cleanup task is tracked [here](https://linear.app/getsentry/issue/PREVENT-168)).

Also note I created a token and added a secret to this repo to enable this feature (`OVERWATCH_SENTRY_AUTH_TOKEN`). This can also get cleaned up with above when ready.

### DIFF
--- a/.github/workflows/bug-prediction.yml
+++ b/.github/workflows/bug-prediction.yml
@@ -1,0 +1,30 @@
+name: Bug Prediction (Overwatch)
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+
+permissions:
+  contents: read
+
+jobs:
+  upload-overwatch:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Overwatch CLI
+        run: |
+          curl -o overwatch-cli https://overwatch.codecov.io/linux/cli
+          chmod +x overwatch-cli
+
+      - name: Run Overwatch CLI
+        run: |
+          ./overwatch-cli \
+            --auth-token ${{ secrets.OVERWATCH_SENTRY_AUTH_TOKEN }} \
+            --organization-slug sentry \
+            --upload-empty-on-error \
+            typescript --package-manager pnpm --eslint-pattern src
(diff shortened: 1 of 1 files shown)


betegon:
hey @suejung-sentry is this still in draft for any reason or can I review it now?

.github/workflows/bug-prediction.yml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Install Overwatch CLI
+        run: |
+          curl -o overwatch-cli https://overwatch.codecov.io/linux/cli
+          chmod +x overwatch-cli
+
+      - name: Run Overwatch CLI
+        run: |
+          ./overwatch-cli \
+            --auth-token ${{ secrets.OVERWATCH_SENTRY_AUTH_TOKEN }} \
```
betegon, line 30: I saw you already added the token

.github/workflows/bug-prediction.yml
```diff
@@ -0,0 +1,30 @@
...
+            --organization-slug sentry \
+            --upload-empty-on-error \
+            typescript --package-manager pnpm --eslint-pattern src
```
betegon, line 33: hrm, we're actually using biome here, not sure if the cli has an option for that

suejung-sentry:
> hey @suejung-sentry is this still in draft for any reason or can I review it now?

@betegon , thanks for pinging. It was in draft while I was fiddling with it, but should be ready for review now. 
I'm debugging why I didn't see a PR comment come in here, but that is independent from this setup.

betegon (changes requested):
Thank you! I want to see it in action :) 
- maybe we can change the eslint comment? Not sure how relevant it is as I imagine it will fire a PR, so we can do formatting later
- Date to remove the step (based on linear project), should be something like september? to keep an eye on it

.github/workflows/bug-prediction.yml
suejung-sentry: Yup! That is getting picked up and working as expected 👍

.github/workflows/bug-prediction.yml
suejung-sentry: This is technically doing a force upload no-op. I'll add in a comment as such so it's clearer

betegon: oh, got it! thanks

PR #346, opened 2025-07-03
wip: add event attachment download support

seanhoughton:
Attachments can be listed and downloaded. Text attachments are returned inline with the text response. Binary attachments are included as embedded resource.

This is a prototype implementation to address my feature request: #314. It works very well for small-ish text attachments. I think we should consider a size limit and just return the download URL for anything larger than that limit.

I'm also wondering of we should detect image types and return them as `type: "image"` content blocks.

I've discovered that VSCode does allow you to download embedded resources, but it's a tiny button that's hidden unless you expand the results of a tool execution. It does not allow you to save the resource to disk as part of the chat.


packages/mcp-server/src/tools.ts
```diff
@@ -457,6 +461,133 @@ export const TOOL_HANDLERS = {
...
+        output += `**Size:** ${attachment.size} bytes\n`;
+        output += `**MIME Type:** ${attachment.mimetype}\n`;
+        output += `**Created:** ${attachment.dateCreated}\n`;
+        output += `**SHA1:** ${attachment.sha1}\n\n`;
+        output += `To download this attachment, use the "get_event_attachment" tool:\n`;
+        output += `\`get_event_attachment(organizationSlug="${params.organizationSlug}", projectSlug="${params.projectSlug}", eventId="${params.eventId}", attachmentId="${attachment.id}")\`\n\n`;
+      });
+
+      return output;
+    } catch (error) {
```
dcramer: id probably not even bother with this - if we really need to try/catch the main query i would wrap it much farther up otherwise it just gets hard to read

that said, given a lot of things can return 404 my intent is just that we spell out generalized error handling for the LLMs at a higher level

packages/mcp-server/src/tools.ts
```diff
@@ -457,6 +461,133 @@ export const TOOL_HANDLERS = {
       apiService,
     });
   },
+  get_event_attachment: async (context, params) => {
+    const apiService = apiServiceFromContext(context, {
+      regionUrl: params.regionUrl,
+    });
+
+    if (!params.organizationSlug) {
```
dcramer: shouldn't need any validation for the params that aren't marked as optional - the function signature is a bit weird if you call it directly, but the tool service itself should be validating them just fine

packages/mcp-server/tsconfig.json
```diff
@@ -1,10 +1,9 @@
 {
-  "extends": "@sentry/mcp-server-tsconfig/tsconfig.base.json",
+  "extends": "../mcp-server-tsconfig/tsconfig.base.json",
```
dcramer: shouldn't need this (and could become invalid) - were you getting an error?

PR #400, opened 2025-07-17
fix: correct user agent field mapping in search_events tool

dcramer:
## Summary

This PR fixes the user agent query mapping in the search_events tool. Previously, queries about "user agents" were incorrectly being mapped to `user.id` fields instead of the correct `user_agent.original` field that contains the actual browser/client string.

### Bug Fix
- Updated AI agent prompts to correctly map "user agents", "browsers", and "clients" to `user_agent.original`
- Added explicit guidance to use the datasetAttributes tool for ambiguous field names
- Prevents incorrect aggregation by user IDs when users ask about browser/client statistics

### Examples Added
- Added user agent query examples for spans dataset (e.g., "which user agents have the most tool calls")
- Added user agent examples for errors dataset (e.g., "errors by browser")
- Demonstrates proper use of `has:user_agent.original` filtering

### Test Coverage
- Added comprehensive test case verifying correct field mapping
- Ensures queries use `user_agent.original` instead of `user.id`

### Documentation
- Updated search-events CLAUDE.md with User Agent Field Mapping section
- Documents the correct field usage for future reference

## Test plan

- All existing tests pass
- New test specifically validates user agent query handling
- TypeScript compilation and linting pass

/cc @claude-code

### DIFF
--- a/packages/mcp-server/src/tools/search-events.test.ts
+++ b/packages/mcp-server/src/tools/search-events.test.ts
@@ -310,12 +310,16 @@ describe("search_events", () => {
 
   it("should correctly handle user agent queries", async () => {
     // Mock AI response for user agent query in spans dataset
-    mockGenerateText.mockResolvedValueOnce(
-      mockAIResponse("spans", "has:user_agent.original", [
-        "user_agent.original",
-        "count()",
-      ]),
-    );
+    mockGenerateText.mockResolvedValueOnce({
+      experimental_output: {
+        dataset: "spans",
+        query: "has:mcp.tool.name AND has:user_agent.original",
+        fields: ["user_agent.original", "count()"],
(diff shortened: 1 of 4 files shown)


dcramer:
Thanks for the review @cursor\! I've addressed the feedback in commit 9c2a35f:

1. ✅ Updated the mocked AI response to include the complete query with `has:mcp.tool.name` filter for "tool calls"
2. ✅ Added the `timeRange` parameter with `statsPeriod: '24h'` for "yesterday"
3. ✅ Fixed the sort expectation to `-count` (without parentheses) to match the API transformation

The test now properly validates the complete user agent query scenario and passes successfully.

dcramer:
alright which robot said that on my behalf

PR #489, opened 2025-08-12
Implement canary deployment strategy with separate workers

dcramer:
## Summary

Implements a safe deployment strategy using separate Cloudflare workers for canary testing before production.

## Changes

- Deploy to `sentry-mcp-canary` worker for testing, then `sentry-mcp` for production
- Run smoke tests against both deployments with automatic rollback on failure
- Add 30-second propagation waits and JUnit test reporting
- Single build, two named worker deployments (not environments)

## Flow

Build → Deploy Canary → Test → Deploy Production → Test → Rollback if needed


dcramer:
I feel like environments are probably the better call here yet, but I need to figure out exactly how they're supposed to work:

> ⏺ Yes, currently we do need to rebuild for each environment because:

>  1. Vite requires environment at build time: When we set CLOUDFLARE_ENV=canary, Vite bakes the environment configuration into the build output
>  2. Resource IDs are embedded: The KV namespace IDs, rate limiter namespace IDs, etc. get compiled into the dist/sentry_mcp/wrangler.json during build

PR #533, opened 2025-09-02
fix(docs) proper redirect uri

sergical:
Was giving an invalid redirect uri error when connecting with `http://localhost:5173/callback` set in the application dashboard, updating to `http://localhost:5173/oauth/callback` fixes it. tested on localhost:5173 and the CF inspector


betegon (approved):
thanks @sergical !

PR #549, opened 2025-09-15
Add denied tools command line option

### LINKED ISSUE #196: Enable or disable list of tools (toolsets)
In some cases I only want to use certain tools or I only want users to have access to certain tools. At least when using Roo Code (not sure how other clients work) each tool you add increases bloat in the system prompt so disabling tools also improves efficiency.

Other MCP servers implement this with command line arguments or environment variables. For example, the Github MCP server calls them "[toolsets](https://github.com/github/github-mcp-server?tab=readme-ov-file#specifying-toolsets)".

colinmollenhour:
Adds tool filtering functionality to reduce context bloat for AI agents. Tools matching a regex pattern are completely hidden from the MCP server response.

Fixes #196 

(authored using Claude Code)

### Key Changes
- Added `SENTRY_DENIED_TOOLS_REGEX` environment variable for tool filtering
- Added `--denied-tools=<regex>` CLI argument with precedence over environment variable
- Updated homepage to dynamically show filtered tools based on server configuration
- Comprehensive error handling for invalid regex patterns
- Updated help text and README with usage examples
- Maintains backward compatibility with existing functionality

### Usage
```bash
# Environment variable approach
export SENTRY_DENIED_TOOLS_REGEX="^search_"
pnpm start:stdio --access-token=TOKEN --host=sentry.io

# Command line argument approach (CLI takes precedence)
pnpm start:stdio --access-token=TOKEN --host=sentry.io --denied-tools="^search_"

# Hide all but a specific set of tools
pnpm start:stdio --access-token=TOKEN --denied-tools="^(?!(?:find_projects|get_issue_details)$).*$"

# Hide specific tools
pnpm start:stdio --access-token=TOKEN --denied-tools="^(whoami|get_doc)$"
```

### Technical Details
- Tools are filtered at the MCP protocol level, ensuring they don't appear in server responses
- Invalid regex patterns log warnings but don't crash the server
- Web interface dynamically reflects filtered tool lists for authenticated users
- Filtering function handles edge cases and provides clear error messages

### DIFF
--- a/packages/mcp-cloudflare/src/client/pages/home.tsx
+++ b/packages/mcp-cloudflare/src/client/pages/home.tsx
@@ -190,7 +190,7 @@ export default function Home({ onChatClick }: HomeProps) {
               const toolsToShow =
                 isAuthenticated && metadata
                   ? TOOL_DEFINITIONS.filter((tool) =>
-                      metadata.tools.includes(tool.name),
+                      (metadata.tools || []).includes(tool.name),
                     )
                   : TOOL_DEFINITIONS;
 


packages/mcp-cloudflare/src/client/pages/home.tsx
```diff
@@ -156,45 +164,89 @@ export default function Home({ onChatClick }: HomeProps) {
...
-                                key={s}
-                                className="inline-flex items-center rounded-full border border-violet-500/40 bg-violet-500/10 px-2 py-0.5 text-xs font-mono text-violet-200"
-                              >
-                                {s}
+            {(() => {
+              // Use filtered tools from metadata if available and authenticated, otherwise fall back to static definitions
+              const toolsToShow =
+                isAuthenticated && metadata
+                  ? TOOL_DEFINITIONS.filter((tool) =>
+                      metadata.tools.includes(tool.name),
```
colinmollenhour:
```suggestion
                      (metadata.tools || []).includes(tool.name),
```

colinmollenhour:
> cannot be replicated with the remote service

I think it could by using a query parameter as I mentioned on #196, and I gave an example (Brightdata). If you want me to implement the query parameter to get this PR merged I would consider it, I was just trying to stick with what I knew best and it sounds like it opens a big can of worms regarding other possible settings.. so for now it seems just adding this to stdio makes sense.

Some anecdote.. Sentry is ahead of the curve in adopting http, but pretty much all of the other MCPs I use are stdio so I guess I don't get what the resistance to keeping and supporting stdio around is all about.. My team and I are all programmers so having npx and copying and pasting an API key is just not a big deal to us. Forking is also possible - I hate forking - but if it saves a few thousand tokens for tools and descriptions that we don't use, that could be worthwhile (I literally just want to get the issue details via MCP).

Anyway, thanks for considering and providing and supporting an excellent MCP for Sentry.

dcramer:
I have an alternative solution to this that I think solves it in a much cleaner way that I'm working on (and works everywhere). Ultimately for our cloud service we really do not want to support people generating api keys and sticking them in some config to make exfiltration easy, so the stdio case for cloud isnt really desirable in the long term.

colinmollenhour:
Thanks, I can just wait until it's live unless you'd prefer I test this use-case first - if I can configure the OpenAI compatible url I should be able to test it pretty easily.

Edit: My prompt just starts out like this:

> Inspect Sentry error #$ARGUMENTS using the Sentry MCP and help me... The Sentry organization slug is "xxx" and you'll need to to get the issue details by it's Short ID.

So I'm just using `search_issues` and `get_issue_details`

PR #551, opened 2025-09-19
feat(docs) add Codex config

sergical:
rest of the changes are from `biome format` and `biome lint:fix`

### DIFF
--- a/packages/mcp-cloudflare/src/client/components/fragments/remote-setup.tsx
+++ b/packages/mcp-cloudflare/src/client/components/fragments/remote-setup.tsx
@@ -24,6 +24,12 @@ export default function RemoteSetup() {
     args: ["-y", `${NPM_REMOTE_NAME}@latest`, endpoint],
   };
 
+  const codexRemoteConfigToml = [
+    "[mcp_servers.sentry]",
+    'command = "npx"',
+    `args = ["-y", "${NPM_REMOTE_NAME}@latest", "${endpoint}"]`,
+  ].join("\n");
+
   const sentryMCPConfig = {
     url: endpoint,
   };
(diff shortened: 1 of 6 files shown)


packages/mcp-cloudflare/src/client/components/fragments/remote-setup.tsx
```diff
@@ -164,6 +170,46 @@ export default function RemoteSetup() {
...
+            </li>
+            <li>
+              Next time you run <code>codex</code>, the Sentry MCP server will
+              be available. It will automatically open the OAuth flow to connect
+              to your Sentry account.
+            </li>
+          </ol>
+          <p>
+            <small>
+              Need verbose logs? Launch Codex with
```
dcramer: What’s the reason for this callout?

sergical: in case there's connection issues with the MCP server? that's how i debugged the fact that it wasn't connecting before stopping and starting `codex` with the mcp server changed being picked up 😅 could remove it tho tbh

dcramer: id remove it - its another thing to maintain and users shouldnt need it. if this is really a common occurrence we should just link to docs but folks gotta be able to figure out how to read e.g. the codex docs on their own

packages/mcp-cloudflare/src/client/components/fragments/stdio-setup.tsx
```diff
@@ -166,6 +173,32 @@ export default function StdioSetup() {
...
+              configuration.
+            </li>
+          </ol>
+          <p>
+            <small>
+              Codex stores session logs under <code>~/.codex/log/</code>. Set
+              <code>RUST_LOG=codex_core=info</code> before launching if you need
+              verbose output while debugging the MCP connection.
+            </small>
+          </p>
```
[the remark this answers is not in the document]
sergical: thanks 🤖 <3

dcramer (approved):

PR #579, opened 2025-10-22
feat(tools): add safety annotations for tool compliance

codyde:
Per feedback from Anthropic on MCP Safety Guidelines and policy to be in their registry, adding  the tool safety annotations as follows - 

* Add readOnlyHint: true to tools that only read data without modification (such as authentication checks, listing operations, searching, and documentation retrieval)
* Add destructiveHint: false to tools that perform only additive updates (such as creating teams, projects, and DSNs)
* Explicitly set destructiveHint: true for tools that modify or update existing data (such as update operations)
* Consider adding idempotentHint: true for operations that can be safely repeated with the same arguments

Ensure all 19 tools have appropriate safety annotations

- Introduced safety annotations to all tools for MCP directory compliance, including `readOnlyHint`, `destructiveHint`, `idempotentHint`, and `openWorldHint`.
- Updated relevant tools to include these annotations, ensuring clear communication of tool behavior and data modification capabilities.
- Added documentation on required annotations and their usage patterns to `docs/adding-tools.mdc`.

### DIFF
--- a/packages/mcp-server/src/api-client/schema.ts
+++ b/packages/mcp-server/src/api-client/schema.ts
@@ -358,6 +358,9 @@ const BaseEventSchema = z.object({
         .passthrough(),
     )
     .optional(),
+  // "context" (singular) is the legacy "extra" field for arbitrary user-defined data
+  // This is different from "contexts" (plural) which are structured contexts
+  context: z.record(z.string(), z.unknown()).optional(),
   tags: z
     .array(
       z.object({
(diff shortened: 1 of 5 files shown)


packages/mcp-server/src/tools/types.ts
```diff
@@ -14,6 +14,12 @@ export interface ToolConfig<
   description: string;
   inputSchema: TSchema;
   requiredScopes: Scope[];
+  annotations?: {
```
dcramer: lets make this required so we dont miss it in future - i think everything will have at least one annotation

codyde: Done!

PR #627, opened 2025-11-06
feat(dev): add devenv for pnpm

joshuarli:
assuming devenv is already installed globally:

```
devenv sync
direnv allow
```

...performs an isolated node+pnpm install independent of the system

### DIFF
--- a/devenv/sync.py
+++ b/devenv/sync.py
@@ -1,0 +1,67 @@
+from __future__ import annotations
+
+import json
+import os
+import shutil
+
+from devenv import constants
+from devenv.lib import config, fs, proc
+
+
+def installed_pnpm(version: str, binroot: str) -> bool:
+    if shutil.which("pnpm", path=binroot) != f"{binroot}/pnpm" or not os.path.exists(
+        f"{binroot}/node-env/bin/pnpm"
+    ):
+        return False
+
+    stdout = proc.run((f"{binroot}/pnpm", "--version"), stdout=True)
+    installed_version = stdout.strip()
+    return version == installed_version
+
+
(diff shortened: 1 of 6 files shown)


devenv/sync.py
```diff
@@ -0,0 +1,67 @@
...
+        cfg["node"][f"{constants.SYSTEM_MACHINE}_sha256"],
+        reporoot,
+    )
+
+    with open(f"{reporoot}/package.json") as f:
+        package_json = json.load(f)
+        pnpm = package_json["packageManager"]
+        pnpm_version = pnpm.split("@")[-1]
+
+    # TODO: move pnpm install into devenv
```
joshuarli, line 64: i've been procrastinating a bit on https://linear.app/getsentry/issue/DI-1012/devenv-install-pnpm but since more repos are using pnpm i'll probably get around to it next week!

joshuarli: will update this PR after https://github.com/getsentry/devenv/pull/210 gets merged + new release!

joshuarli:
it automatically installs the pinned pnpm version in `packageManager` (used by node's corepack but corepack is deprecated now, also used by https://github.com/pnpm/action-setup)

otherwise you can install and manage pnpm yourself but letting devenv handle it helps us ensure consistency and repeatabilty across dev+ci environments

dcramer:
I'm going to say we wont do this. Happy to make it optional (right now it basically makes it the default way), but devenv is a global installation on machines and we're already using industry standard dev environment practices for this repo. I dont want to require third party contributors to setup a Sentry specific tool when its not needed.

PR #635, opened 2025-11-14
fix(openai-provider): only apply reasoning effort to o1/o3 models

ZhiRuYan:
## Summary

Fix OpenAI provider to only apply the `reasoningEffort` parameter to models that support it (o1 and o3 series). Previously, this parameter was applied to all models when `OPENAI_REASONING_EFFORT` was set, which could cause unnecessary overhead or potential errors with non-o1/o3 models like gpt-5.

## Changes

- Added `supportsReasoningEffort()` helper function to check if a model supports the reasoning effort parameter
- Modified logic to only apply reasoning effort to o1/o3 models
- Updated default behavior: reasoning effort defaults to "low" only for o1/o3 models, not for all models
- Improved code comments to clarify the behavior
- Fixed variable reuse by introducing `selectedModel` variable for better code clarity

## Motivation

The reasoning effort parameter is specific to OpenAI's o1 and o3 model series. Applying it to other models (like gpt-5, gpt-4, etc.) is unnecessary and could potentially cause issues.

## Testing

- ✅ All existing tests pass
- ✅ TypeScript compilation successful
- ✅ Linting passes
- ✅ Code follows repository patterns

### DIFF
--- a/packages/mcp-server/src/internal/agents/openai-provider.ts
+++ b/packages/mcp-server/src/internal/agents/openai-provider.ts
@@ -22,33 +22,46 @@ export function setOpenAIBaseUrl(baseUrl: string | undefined): void {
 }
 
 /**
+ * Check if a model supports reasoning effort parameter (o1 series only)
+ */
+function supportsReasoningEffort(modelName: string): boolean {
+  return modelName.startsWith("o1-") || modelName.startsWith("o3-");
+}
+
+/**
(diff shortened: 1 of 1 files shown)


dcramer:
Closing after review. Restricting reasoning effort to only o1/o3 has an invalid model assumption and would exclude newer reasoning-capable models. The provider implementation has also moved to the centralized provider architecture.

PR #660, opened 2025-11-25
feat(mcp-cloudflare): Make our mcp server stateful

betegon:
This will allow us to:
- have an `session_id` for each session, so we can track a full session in Sentry. We will know the average number of tool calls per session for example.
- track the client name and transport of each session.

### DIFF
--- a/packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
+++ b/packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
@@ -102,6 +102,14 @@ const mcpHandler: ExportedHandler<Env> = {
         });
       }
       expandedScopes = Array.from(expandScopes(new Set(valid)));
+
+      // Validate that at least one valid scope was granted (matches skills validation)
+      if (valid.size === 0) {
+        return new Response(
+          "Authorization failed: No valid scopes were granted. Please re-authorize and select at least one permission.",
+          { status: 400 },
+        );
+      }
     }
 
     // Parse and validate granted skills (NEW - primary authorization method)


packages/mcp-cloudflare/wrangler.jsonc
```diff
@@ -65,12 +65,24 @@
...
   "migrations": [
     {
       "tag": "v1"
     },
     {
       "tag": "v2"
+    },
+    {
+      "tag": "v3",
+      "new_sqlite_classes": ["McpSession"]
```
betegon, line 85: `McpSession` extends `Agent` and it requires this as its backing store

packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
```diff
@@ -87,7 +83,7 @@ const mcpHandler: ExportedHandler<Env> = {
     }
 
     // Parse and expand granted scopes (LEGACY - for backward compatibility)
-    let expandedScopes: Set<Scope> | undefined;
+    let expandedScopes: string[] | undefined;
```
betegon, line 91: as we're having these in the durable object, we need to make them a string array as when we call a durable object from a worker, the data is serialized by RPC and it doesn't support `Set`.

dcramer: actually we should add that as a comment here mostly so LLMs dont fuck it up in the future

packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
```diff
@@ -150,35 +146,34 @@ const mcpHandler: ExportedHandler<Env> = {
...
       // Scopes derived from skills - for backward compatibility with old MCP clients
       // that don't support grantedSkills and only understand grantedScopes
       grantedScopes: expandedScopes,
       grantedSkills, // Primary authorization method
-      constraints: verification.constraints,
+      organizationSlug: verification.constraints.organizationSlug,
+      projectSlug: verification.constraints.projectSlug,
       sentryHost,
       mcpUrl: env.MCP_URL,
+      isAgentMode,
```
dcramer, line 179: is this a bug or is the diff just looking odd

packages/mcp-cloudflare/src/server/session.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      // Update session context with fresh data from this request
+      this.sessionContext.userId = contextData.userId;
+      this.sessionContext.clientId = contextData.clientId;
+      this.sessionContext.accessToken = contextData.accessToken;
+      this.sessionContext.sentryHost = contextData.sentryHost;
+      this.sessionContext.mcpUrl = contextData.mcpUrl;
+
+      if (contextData.grantedScopes) {
+        this.sessionContext.grantedScopes = new Set<Scope>(
```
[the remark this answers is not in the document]
betegon: The only falsy value it could be is undefined, which is the intended

packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
betegon: not a bug, we pass all request-derived config through `SerializableServerContext` so we keep the URL parsing in the edge handler.

packages/mcp-cloudflare/src/server/session.ts
```diff
@@ -0,0 +1,144 @@
...
+      // Context is captured in tool handler closures during buildServer()
+      this.server = buildServer({
+        context: this.sessionContext,
+        agentMode,
+        onToolComplete: () => {
+          // Flush Sentry events after tool execution
+          this.ctx.waitUntil(Sentry.flush(2000));
+        },
+      });
+    }
```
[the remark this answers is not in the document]
betegon: auth won't change between requests of the same session. Flow is:
1. user authenticates with X permissions
2. user does some tool call
3. user reauthenticate for whatever reason. e.g. adding more permissions
4. new session is created, thus scopes/skills are set at that point in the new DO for the new session.

packages/mcp-cloudflare/src/server/session.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (contextData.grantedScopes) {
+        this.sessionContext.grantedScopes = new Set<Scope>(
+          contextData.grantedScopes as Scope[],
+        );
+      }
+
+      if (contextData.grantedSkills) {
+        this.sessionContext.grantedSkills = new Set<Skill>(
+          contextData.grantedSkills as Skill[],
+        );
```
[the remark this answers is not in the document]
betegon: scopes are granted with the oauth flow so they are granted during all the session but changed it to be more defensive.

packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
```diff
@@ -100,11 +101,19 @@ const mcpHandler: ExportedHandler<Env> = {
           },
         });
       }
-      expandedScopes = expandScopes(new Set(valid));
+      expandedScopes = Array.from(expandScopes(new Set(valid)));
+
+      // Validate that at least one valid scope was granted (matches skills validation)
+      if (valid.size === 0) {
```
dcramer, line 107: this ones actually wrong - scopes are deprecated in favor of skills. basically scopes only exist with legacy session data and are getting phased out entirely

PR #667, opened 2025-12-03
feat: enable telemetry for embedded AI agents

dcramer:
Enable input/output recording in Vercel AI integration and add experimental telemetry to callEmbeddedAgent for better observability of AI-powered tool calls.

### DIFF
--- a/packages/mcp-core/src/internal/agents/callEmbeddedAgent.ts
+++ b/packages/mcp-core/src/internal/agents/callEmbeddedAgent.ts
@@ -45,6 +45,10 @@ export async function callEmbeddedAgent<
     maxSteps: 5,
     temperature: 1, // GPT-5 only supports temperature of 1
     experimental_output: Output.object({ schema }),
+    experimental_telemetry: {
+      isEnabled: true,
+      functionId: "callEmbeddedAgent",
+    },
     // Disable strict schema validation for both output and tool parameter schemas.
     //
     // OpenAI's structured outputs have limitations:


packages/mcp-cloudflare/src/server/sentry.config.ts
```diff
@@ -27,7 +27,10 @@ export default function getSentryConfig(env: Env): SentryConfig {
     integrations: [
       Sentry.consoleLoggingIntegration(),
       Sentry.zodErrorsIntegration(),
-      Sentry.vercelAIIntegration(),
+      Sentry.vercelAIIntegration({
```
dcramer: pretty sure it was broken last time we tried to add these options but doesnt seem to error now so must be fixed upstream?

packages/mcp-core/src/internal/agents/callEmbeddedAgent.ts
```diff
@@ -45,6 +45,10 @@ export async function callEmbeddedAgent<
     maxSteps: 5,
     temperature: 1, // GPT-5 only supports temperature of 1
     experimental_output: Output.object({ schema }),
+    experimental_telemetry: {
```
dcramer, line 48: docs dont say we need this but its not working so who knows

packages/mcp-cloudflare/src/server/sentry.config.ts
dcramer: jk this still doesnt work

PR #682, opened 2025-12-15
feat: Improve performance queries on event search

dcramer:
## Summary
Enhances the `search_events` tool to better handle performance-related queries by adding explicit guidance for web vitals, latency analysis, and aggregate-first investigation strategies.

### Key Changes
- Added performance investigation strategy to system prompt - ensures queries about "slow pages" or "latency issues" use aggregate queries grouped by transaction instead of misleading individual samples
- Added Web Vitals fields (LCP, CLS, INP, FCP, TTFB) to spans dataset with threshold documentation
- Optimized dataset examples for more concise token usage

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/packages/mcp-core/src/tools/search-events/config.ts
+++ b/packages/mcp-core/src/tools/search-events/config.ts
@@ -357,363 +357,223 @@ export const DATASET_FIELDS = {
   },
 };
 
-// Dataset-specific rules and examples
-export const DATASET_CONFIGS = {
-  errors: {
-    rules: `- For errors, focus on: message, level, error.type, error.handled
-- Use level field for severity (error, warning, info, debug)
-- Use error.handled:false for unhandled exceptions/crashes
-- For filename searches: Use stack.filename for suffix-based search (e.g., stack.filename:"**/index.js" or stack.filename:"**/components/Button.tsx")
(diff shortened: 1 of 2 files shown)


packages/mcp-core/src/tools/search-events/config.ts
```diff
@@ -105,6 +105,31 @@ When user asks for calculated metrics, ratios, or conversions:
 - IMPORTANT: Equations are ONLY supported in the spans dataset, NOT in errors or logs
 - IMPORTANT: When sorting by equations, use "-equation|..." for descending order (highest values first)
 
+PERFORMANCE INVESTIGATION STRATEGY:
+When users ask about "performance problems", "slow pages", "slow endpoints", "latency issues",
```
bcoe, line 109: We may want to extend this to include mobile questions in the future, TTID, ANR.

@noahsmartin might have a strong opinion as to what the right ones are.

I also wonder about other performance domains, e.g., query performance.

I believe starting with page / vitals is a good starting point though.

dcramer: it actually has an endpoint it can query - so it should be able to figure these out

PR #702, opened 2026-01-04
feat: docker-compose environment with streamable-http + oauth2.1 and zero cloudflare dependency

owengo:
The idea is to allow deployment of the MCP server in a container environment ( k8s for example, helm chart coming later ) with oauth 2.1 support without relying on cloudflare. Support for external storage with the keys with valkey so that you can have several instances running redundantly, with encryption ( tokens are encrypted before being sent to redis/valkey ).

This mean you can for example run it  for your shared on-prem librechat environment with your on-prem sentry server, everything stays in-house.

The whole code is generated by Claude ( Opus 4.5 ) and does not modify anything in the main project, it's a new folder with the code to run the dockerized MCP server.

### DIFF
--- a/fastmcp-http-oauth/AGENTS.md
+++ b/fastmcp-http-oauth/AGENTS.md
@@ -56,7 +56,9 @@ fastmcp-http-oauth/
 ├── .env.example           # Environment template
 ├── .env                   # Local environment (gitignored)
 ├── custom-cas/            # Custom CA certificates (for self-signed certs)
-└── __tests__/             # Unit tests
+├── __tests__/             # Unit tests
+└── helm/                  # Helm chart for Kubernetes deployment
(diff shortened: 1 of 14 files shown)


owengo:
> Thank you for the PR!
> 
> I dont think we want to land this in main unfortunately. It would require us to continue to support this and we generally avoid that as a rule of thumb. Its also worth noting the MCP is already fairly brittle when it comes to what it supports for self-hosted (e.g. Seer doesnt exist). We're going to continue moving functionality into Sentry's core to ease some of that burden, like the work on updating the provider to 2.1 in [getsentry/sentry#99002](https://github.com/getsentry/sentry/issues/99002)).

I completely understand, actually I should make a different project which just imports your sentry-mcp node module , it was just quick to do it like this but it would be a burden for your project

PR #733, opened 2026-01-23
feat(tools): Add unified get_profile tool with URL parsing

dcramer:
## Summary
- Add experimental `get_profile` tool for CPU profiling analysis
- Analyze transaction profiles to identify performance bottlenecks
- Compare performance between time periods to detect regressions
- Extract hot paths, performance percentiles (p75/p95/p99), and actionable recommendations

## Changes
- New `get_profile` tool (behind experimental flag)
- Flamegraph API client integration with schema validation
- Profile analyzer for hot path identification and performance insights
- Profile formatter for human-readable output with optimization recommendations
- URL parsing helpers for Sentry profile URLs
- Comprehensive test coverage with inline snapshots

## Test plan
- [x] Unit tests for analyzer, formatter, and URL helpers
- [x] Integration tests for get_profile tool
- [x] Manual verification via CLI with dev server
- [x] Verified experimental flag hides tool by default


packages/mcp-core/src/tools/profile/analyzer.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  const frames = flamegraph.shared.frames;
+  const frameInfos = flamegraph.shared.frame_infos;
+
+  // Calculate total samples for percentage calculation
+  const totalSamples = profile.sample_counts.reduce((a, b) => a + b, 0);
+
+  // Combine samples with their stats
+  const hotPaths = profile.samples.map((stackIndices, idx) => {
```
JonasBa: It is very likely that the same stack will appear multiple times in the samples (same code executed at different points), so you need to do merging after the samples are resolved.

Another reason why you want to merge is that you can have two stack like

foo, bar, baz
foo, bar

In which case the total weight of the bar frame is 2, and the self weight is 1.

packages/mcp-core/src/tools/profile/analyzer.ts
```diff
@@ -0,0 +1,417 @@
...
+    const sampleCount = profile.sample_counts[idx] || 0;
+    const duration = profile.sample_durations_ns[idx] || 0;
+    const weight = profile.weights[idx] || 0;
+
+    // Reconstruct call stack preserving order
+    const callStack = reconstructCallStack(stackIndices, frames, frameInfos);
+
+    // Filter to user code if requested
+    const userCodeFrames = options.focusOnUserCode
```
JonasBa: I would opt to keep all frames in the code, and omit this during formatting if you have to

packages/mcp-core/src/tools/profile/analyzer.ts
```diff
@@ -0,0 +1,417 @@
...
+      count: 0,
+      weight: 0,
+      sumDuration: 0,
+      sumSelfTime: 0,
+      p75Duration: 0,
+      p95Duration: 0,
+      p99Duration: 0,
+    },
+    depth,
+    isRoot: depth === 0,
```
JonasBa: > Frames in a stack should be ordered from leaf to root. It means your main frame should be at the end of the list.

The idea is here is that bc building a flamegraph or call tree requires merging the stacks, you would typically want work top down, so you can break or merge without having to read the entire stack sample

Issue #780, opened 2026-02-06
Unable to authenticate with Sentry MCP server


rodolfoBee:
### Environment
Version: 2.4.28 (Universal)
VSCode Version: 1.105.1

### Steps to Reproduce
1. Remove any Sentry MCP Server configuration if already installed
2. Open mcp.json and add the Sentry server:
```
{
  "mcpServers": {
    "Sentry": {
    "url": "https://mcp.sentry.dev/mcp"
    }
  }
}
```
3. In Cursor Settings -> Tools & MCP try to connect to Sentry.

### Expected Result
The browser opens automatically to authenticate with Sentry. 

### Actual Result
It fails silently. When checking the logs on the Cursor's Output pannel I get the following log:
```
2026-02-06 11:03:33.378 [info] Handling LogoutServer action
2026-02-06 11:03:33.378 [info] Clearing stored OAuth data
2026-02-06 11:03:33.388 [info] Successfully cleared OAuth tokens
2026-02-06 11:03:33.388 [info] Cleaning up
2026-02-06 11:03:33.392 [info] Handling ListOfferings action, server stored: false
2026-02-06 11:03:33.392 [info] Server not yet created, returning empty offerings
2026-02-06 11:03:33.400 [info] Handling ReloadClient action
2026-02-06 11:03:33.401 [info] Creating streamableHttp transport
2026-02-06 11:03:33.408 [info] Handling GetInstructions action
2026-02-06 11:03:33.408 [info] Server creation in progress for getInstructions, waiting
2026-02-06 11:03:33.454 [info] Using OAuth scopes: org:read, project:write, team:write, event:write
2026-02-06 11:03:33.454 [info] Connecting to streamableHttp server
2026-02-06 11:03:33.455 [info] No stored tokens found
2026-02-06 11:03:33.576 [info] No stored client information found
2026-02-06 11:03:33.578 [info] Using redirect URL {"url":"cursor://anysphere.cursor-mcp/oauth/callback"}
2026-02-06 11:03:33.680 [info] Saving client information {"redirects":1,"clientIdPresent":true}
2026-02-06 11:03:33.682 [info] Using redirect URL {"url":"cursor://anysphere.cursor-mcp/oauth/callback"}
2026-02-06 11:03:33.683 [info] No stored tokens found
2026-02-06 11:03:33.683 [info] Using redirect URL {"url":"cursor://anysphere.cursor-mcp/oauth/callback"}
[...]

PR #786, opened 2026-02-06
fix(cloudflare): Add resource_metadata to OAuth 401 response

dcramer:
## Summary

- Adds `resource_metadata` parameter to the `WWW-Authenticate` header in 401 responses per [RFC 9728](https://www.rfc-editor.org/rfc/rfc9728), giving MCP clients a direct pointer to the OAuth protected resource metadata endpoint
- Eliminates one discovery round-trip for clients that use the `resource_metadata` hint
- URL is constructed dynamically to support `/mcp`, `/mcp/:org`, and `/mcp/:org/:project` paths

Refs #780

## Context

Cursor IDE fails to complete the MCP OAuth flow — the root cause is a Cursor client bug (it doesn't wait for the browser callback). While this change won't fix Cursor, it makes our 401 response spec-compliant and may help better-behaved clients.

## Test plan

- [x] `pnpm run tsc && pnpm run lint && pnpm run test` passes
- [ ] `curl -s -D - https://mcp.sentry.dev/mcp -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"initialize","id":1}'` — verify `WWW-Authenticate` header includes `resource_metadata`
- [ ] Authenticated requests still work: `pnpm -w run cli "who am I?"`

### DIFF
--- a/packages/mcp-cloudflare/src/server/index.ts
+++ b/packages/mcp-cloudflare/src/server/index.ts
@@ -91,8 +91,10 @@ const wrappedOAuthProvider = {
         const resourceMetadataUrl = `${url.origin}/.well-known/oauth-protected-resource${url.pathname}`;
         const existingAuth =
           headers["WWW-Authenticate"] ?? headers["www-authenticate"] ?? "";
-        const newAuth = existingAuth
-          ? `resource_metadata="${resourceMetadataUrl}", ${existingAuth}`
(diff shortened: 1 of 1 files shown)


dcramer:
arguably this patch doesnt matter, so not even sure i care to land it at this point as it just adds more noise to the codebase

PR #814, opened 2026-02-25
perf: Parallelize HTTP calls in get_issue_details

mrduncan:
Both code paths (eventId and issueId/issueUrl) previously made 4 sequential HTTP requests after resolving the issue. Now the event fetch (+ its optional trace fetch) runs in parallel with the enrichment data fetch (autofix state + external issues), cutting the sequential waterfall from 4 round-trips to 2-3.

[Trace example](https://sentry.sentry.io/issues/trace/88fd5814146a46cba566ab7fca21d3c9/?groupId=7278843067&node=span-89d4633dfd1422e1&pageEnd=2026-02-20T21%3A09%3A32.859&pageStart=2026-02-19T21%3A09%3A32.859&project=4509062593708032&source=issue_details&timestamp=1771578571) from https://sentry.sentry.io/issues/7278843067/?project=4509062593708032

Fixes MCP-SERVER-F5N

### DIFF
--- a/packages/mcp-core/src/tools/get-issue-details.ts
+++ b/packages/mcp-core/src/tools/get-issue-details.ts
@@ -118,36 +118,39 @@ export default defineTool({
         issue = found;
       }
       // For this call, we might want to provide context if it fails
-      let event: Awaited<ReturnType<typeof apiService.getEventForIssue>>;
-      try {
-        event = await apiService.getEventForIssue({
-          organizationSlug: orgSlug,
-          issueId: issue.shortId,
-          eventId,
-        });
-      } catch (error) {
-        // Optionally enhance 404 errors with parameter context
-        if (error instanceof ApiNotFoundError) {
-          throw enhanceNotFoundError(error, {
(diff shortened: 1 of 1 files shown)


packages/mcp-core/src/tools/get-issue-details.ts
```diff
@@ -200,22 +203,27 @@ export default defineTool({
...
-      event,
-    });
+    const [{ event, performanceTrace }, { autofixState, externalIssues }] =
+      await Promise.all([
+        apiService
+          .getLatestEventForIssue({
+            organizationSlug: orgSlug,
+            issueId: issue.shortId,
+          })
+          .then(async (event) => ({
```
dcramer, line 213: wonder if thi can be done in parallel too? i dont remember exactly what all this code is doing

mrduncan: As far as I can tell, maybeFetchPerformanceTrace needs the trace id from that event result

dcramer: ah sorry i totally missed the event argument being passed in

PR #922, opened 2026-04-26
fix(mcp-core): route structured logs to stderr

3000-2:
Route the LogTape console sink through stderr for all log levels.

This fixes stdio transport disconnects caused by structured LogTape records being written to stdout. MCP stdio transports reserve stdout for JSON-RPC messages, but LogTape's default console level map sends `info` and `debug` records through `console.info` / `console.debug`; in Node.js those write to stdout. When a structured log line lands on stdout, stdio clients can parse it as a protocol message, fail validation because it has fields like `@timestamp`, `level`, `message`, `logger`, and `properties` instead of `jsonrpc`, `id`, or `method`, then close the transport.

This keeps the log record severity in the structured payload, e.g. `"level": "INFO"`. The change only controls which console method LogTape uses for the console sink, so structured logs stay off stdout.

Added a regression test asserting `logInfo` does not write through `console.info`.

### DIFF
--- a/packages/mcp-core/src/telem/logging.test.ts
+++ b/packages/mcp-core/src/telem/logging.test.ts
@@ -1,0 +1,39 @@
+import { describe, expect, it, vi, afterEach } from "vitest";
+
+describe("logging", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    vi.resetModules();
+  });
+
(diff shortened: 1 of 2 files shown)


dcramer:
Damn idk how I missed this. Thanks!

PR #929, opened 2026-04-28
telem(cloudflare): promote User-Agent to a Sentry tag

JoshFerge:
## Summary

- The User-Agent header is already captured on every event as part of `request.headers` (a context), but Sentry contexts aren't tag-indexed — that blocks group-by / filter on User-Agent in the aggregate UI (Issues, Transactions, Discover).
- Set `user_agent` as a tag on the per-request isolation scope created by `Sentry.withSentry`. This applies to both error and transaction events for that request without needing `beforeSendTransaction`.

## Why a tag and not the existing `client_family` bucket

`client_family` (set as a metric attribute) is intentionally low-cardinality. The raw User-Agent is what's needed to investigate one-off clients (e.g. specific Cursor / IDE versions hitting an edge case) which the bucketed family hides.

## Test plan

- [ ] Deploy to staging and confirm the `user_agent` tag appears on transactions hitting `/mcp` and `/oauth/*`
- [ ] Confirm tag shows up in Issues filter dropdown and Discover
- [ ] Verify `pnpm run tsc && pnpm run lint && pnpm run test` (passing locally)

### DIFF
--- a/packages/mcp-cloudflare/src/server/index.ts
+++ b/packages/mcp-cloudflare/src/server/index.ts
@@ -80,10 +80,6 @@ function finalizeResponse(
 //      - Public metadata endpoints → restrictive read-only CORS (`*`, GET only)
 //      - Everything else → strip all CORS headers the library added
 
-// Sentry tag values are truncated server-side to 200 chars; do it here so
-// the value the Sentry UI shows matches what we put on the scope.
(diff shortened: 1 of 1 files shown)


dcramer:
gonna just close since sibling spans have all the request metadata indexed and unfortunately user-agent is also not the most useful tag on mcp sessions (you want the mcp client identifier, which isnt always in the user-agent)

PR #962, opened 2026-05-13
chore(evals): unblock ToolPredictionScorer

JoshFerge:
## Summary
Two unrelated bugs were both blocking the entire eval suite — every eval currently crashes with \`MCPClientError: Connection closed\` before any scoring runs. Both are fixed here.

1. **Stale CLI flag.** \`toolPredictionScorer\` spawns \`pnpm exec sentry-mcp\` with \`--access-token=mocked-access-token --all-scopes\`. \`--all-scopes\` was removed in #668 ("remove legacy grantedScopes in favor of grantedSkills"), so the spawned binary prints its usage banner and exits, surfacing as a closed-stdio error. Drop the flag.
2. **OpenAI strict response_format rejection.** The prediction schema uses \`arguments: z.record(z.any()).optional().default({})\`. zod-to-JSON-Schema emits this as \`additionalProperties: {}\` (no \`type\` key), which OpenAI's strict structured-output endpoint refuses with \`AI_APICallError: Invalid schema for response_format 'response': … schema must have a 'type' key\`. Verified this also fails on the latest \`@ai-sdk/openai@3.0.63\` + \`ai@6.0.182\`, so it's not a stale-dep issue. Encode arguments as a JSON string instead.

With both fixed, the eval suite runs end-to-end again (verified locally — autofix.eval.ts cases score 1.00 against the current main).

## Test plan
- [x] \`pnpm run tsc\`
- [x] \`pnpm run lint\`
- [x] \`pnpm run test\` — all 1341 tests pass
- [x] \`pnpm vitest run autofix.eval.ts\` with \`OPENAI_API_KEY\` set — both cases score 1.00

## Notes for reviewers
- \`predictedTools[*].arguments\` was already only read for the rationale/metadata field in the scorer's return value, so renaming to \`argumentsJson\` doesn't break scoring.
- First in a stack of three. Subsequent PRs do the Sentry-side autofix migration to explorer-mode endpoints.

Agent transcript: https://claudescope.sentry.dev/share/HYSozfr18sU5lMEbTaUO5UX2YmVyaBdNoueQpOS6xJQ

### DIFF
--- a/packages/mcp-server-evals/src/evals/utils/toolPredictionScorer.ts
+++ b/packages/mcp-server-evals/src/evals/utils/toolPredictionScorer.ts
@@ -19,12 +19,7 @@ async function getAvailableTools(): Promise<string[]> {
   // Use pnpm exec to run the binary from the workspace
   const transport = new Experimental_StdioMCPTransport({
     command: "pnpm",
-    args: [
-      "exec",
-      "sentry-mcp",
-      "--access-token=mocked-access-token",
-      "--all-scopes",
-    ],
+    args: ["exec", "sentry-mcp", "--access-token=mocked-access-token"],
     env: {
       ...process.env,
       SENTRY_ACCESS_TOKEN: "mocked-access-token",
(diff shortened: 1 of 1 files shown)


packages/mcp-server-evals/src/evals/utils/toolPredictionScorer.ts
```diff
@@ -19,12 +19,7 @@ async function getAvailableTools(): Promise<string[]> {
   // Use pnpm exec to run the binary from the workspace
   const transport = new Experimental_StdioMCPTransport({
     command: "pnpm",
-    args: [
-      "exec",
-      "sentry-mcp",
-      "--access-token=mocked-access-token",
-      "--all-scopes",
```
dcramer, line 26: im actually surprised this wouldnt break something

PR #995, opened 2026-05-21
feat(mcp-core): Forward DCR client identity to Sentry API

yuvmen:
## Summary
- Forward DCR-registered `client_id` and `client_name` as `X-Sentry-MCP-Client-Id` and `X-Sentry-MCP-Client-Name` headers on upstream API requests
- Persist `clientName` from OAuth callback into WorkerProps so it's available per-request
- Enables Sentry to identify which MCP client (e.g. Claude Code, Cursor) made a given API call — not possible today since all MCP traffic uses the same OAuth app

## Test plan
- [ ] Verify `X-Sentry-MCP-Client-Id` and `X-Sentry-MCP-Client-Name` headers appear on outgoing API requests
- [ ] Verify no headers sent when values are null/undefined (e.g. stdio transport)
- [ ] Verify existing tokens without `clientName` in props still work (backward compat)

### DIFF
--- a/packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
+++ b/packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
@@ -157,6 +157,7 @@ const mcpHandler: ExportedHandler<Env> = {
     const userId = rawProps.id as string;
     const accessToken = rawProps.accessToken as string;
     const clientId = rawProps.clientId as string;
+    const clientName = rawProps.clientName;
     const sentryHost = env.SENTRY_HOST || "sentry.io";
     const clientFamily = resolveClientFamily(request.headers.get("user-agent"));
     const requestGrantId = getRequestGrantId(request);
@@ -326,7 +327,7 @@ const mcpHandler: ExportedHandler<Env> = {
       userId,
       userIpAddress,
       clientId,
-      clientFamily,
+      clientName,
       accessToken,
       grantedSkills: validSkills,
       constraints,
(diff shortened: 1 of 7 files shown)


dcramer:
It might make more sense just to pass along the actual client information vs arbitrary header things

e.g. forward the details from DCR as headers (client ID and name?)

mostly because if we need this information in analytics we should future proof this in case we move DCR upstream, and we probably already want this to audit abuse etc

yuvmen:
yea I understand what you are saying, I thought its useful though to use the standardization the MCP layer gives over the `client-name` right? otherwise Sentry will have to deal with it, which it can, but I thought it makes sense for it to resolved at the MCP layer. Not sure how to future proof it though
Fully agree on the client ID though, will pass that as well

dcramer:
actually looking more now: user-agent is already unreliable so im talking about something entirely different

when you do DCR you register oauth details. we should just grab those details and pass them along

user-agent for example historically just shows 'node' or something else for folks because it wasnt set, but the client ID and the client Name (which is user provided) are really what identify the application

basically we should treat MCP as no different than any other oauth client here, and tbqh its an oversight that im not already forwarding this information upstream since its a pita to track for us without it (wrt to e.g. security etc)

yuvmen:
alright sounds good, swapped it to send them

dcramer (approved):
i havent touched this iin a while but went through the code paths and i THINK the server context will "just work" but im suspect based on some historical stuff i remember

mostly we'll wanna validate in prod once this goes out that we have the right information on all requests vs specific ones

PR #996, opened 2026-05-21
fix: Set app.client.family as Sentry attribute for Explore indexing

wedamija:
The \`app.client.family\` bucketing (claude-code, cursor, claude-desktop, etc.) was being set as an attribute on specific metrics but not on spans, making it unavailable for groupBy in Sentry Explore's trace view.

This adds \`Sentry.getActiveSpan()?.setAttribute("app.client.family", clientFamily)\` in both the root request handler (\`index.ts\`) and the MCP request handler (\`mcp-handler.ts\`), covering all request paths where a span is active.

Note: \`tokenExchangeCallback\` in \`oauth/helpers.ts\` does not do this because \`getActiveSpan()\` is \`undefined\` in that callback context — it intentionally uses metric attributes only.

### DIFF
--- a/packages/mcp-cloudflare/src/server/index.ts
+++ b/packages/mcp-cloudflare/src/server/index.ts
@@ -170,6 +170,7 @@ const wrappedOAuthProvider = {
     }
 
     const clientFamily = resolveClientFamily(request.headers.get("user-agent"));
+    Sentry.getActiveSpan()?.setAttribute("app.client.family", clientFamily);
 
     // --- Phase 2: Let the OAuth library handle the request ---
     // We normalize any CORS headers it returns in the response handling below.
--- a/packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
+++ b/packages/mcp-cloudflare/src/server/lib/mcp-handler.ts
@@ -166,6 +166,8 @@ const mcpHandler: ExportedHandler<Env> = {
       userId,
     );
 
+    Sentry.getActiveSpan()?.setAttribute("app.client.family", clientFamily);
+
     // Parse and validate granted skills (primary authorization method)
     // Legacy tokens without grantedSkills are no longer supported
     if (!rawProps.grantedSkills) {


dcramer:
you cant group by on non-tags?

https://sentry.sentry.io/explore/traces/?aggregateField=%7B%22groupBy%22%3A%22mcp.client.name%22%7D&aggregateField=%7B%22yAxes%22%3A%5B%22count%28span.duration%29%22%5D%7D&field=id&field=span.name&field=span.description&field=span.duration&field=transaction&field=timestamp&field=mcp.client.name&mode=aggregate&project=4509062593708032&statsPeriod=30d

wedamija:
> you cant group by on non-tags?
> 
> https://sentry.sentry.io/explore/traces/?aggregateField=%7B%22groupBy%22%3A%22mcp.client.name%22%7D&aggregateField=%7B%22yAxes%22%3A%5B%22count%28span.duration%29%22%5D%7D&field=id&field=span.name&field=span.description&field=span.duration&field=transaction&field=timestamp&field=mcp.client.name&mode=aggregate&project=4509062593708032&statsPeriod=30d

I was specifically trying to group on family, since it's less granular: https://github.com/getsentry/sentry-mcp/blob/6dd16b74bcebb4b07f8f7093512f1b90b6274d48/packages/mcp-cloudflare/src/server/lib/client-family.ts#L14-L34

The llm generated reason might be wrong here. But I can't group by family at the moment in the ui

dcramer:
something must be wrong you should be able to group on any attribute 🤔 

we dont use tags for basically anything outside of some core issue stuff

we're talking spans right?

wedamija:
> something must be wrong you should be able to group on any attribute 🤔
> 
> we dont use tags for basically anything outside of some core issue stuff
> 
> we're talking spans right?

Ahh, ok. It's not even setting this as an attribute, I just trusted what Claude said too much. Switched over to attributes instead of tags

PR #998, opened 2026-05-23
fix(resource): Extract query params from AI conversation URLs

sergical:
## Summary
- Parse `project`, `spanId`, `start`, and `end` query params from AI conversation URLs (e.g. `https://sentry.sentry.io/explore/conversations/slack%3A.../?project=4510944073809921&spanId=...&start=...&end=...`)
- Thread the project ID through `get_sentry_resource` → `get_ai_conversation_details` so the Snuba query scopes to the correct project instead of scanning all projects (`-1`)
- Decode URL-encoded conversation IDs (e.g. `slack%3A...` → `slack:...`)
- Follows the existing pattern used by profile URLs for query param extraction

## Test plan
- [x] Added test for conversation URL with all query params (project, spanId, start, end)
- [x] Existing tests for conversation URLs without query params still pass (return `undefined` for optional fields, matching profile pattern)
- [x] All 1,413 tests pass (`pnpm run test --force`, 0 cached)

### DIFF
--- a/packages/mcp-core/src/internal/url-helpers.test.ts
+++ b/packages/mcp-core/src/internal/url-helpers.test.ts
@@ -241,7 +241,11 @@ describe("parseSentryUrl", () => {
       ).toMatchInlineSnapshot(`
         {
           "conversationId": "conv-123",
+          "end": undefined,
           "organizationSlug": "my-org",
+          "projectSlug": undefined,
+          "spanId": undefined,
+          "start": undefined,
           "type": "ai_conversation",
         }
       `);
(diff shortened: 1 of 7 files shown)


sergical:
Addressed both Warden findings in 51787d2:

**R3W-65G** (URL `?project=` bypasses constraint): Flipped priority — `context.constraints.projectSlug` now always takes precedence over the URL-supplied `params.project`. The constraint is resolved to a numeric ID via `apiService.getProject`; the URL param is only used when no constraint exists.

**GAS-XGJ** (slug passed as numeric ID): Removed the `params.projectSlug` fallback in `get-sentry-resource` so the constraint slug never leaks through to the handler's `project` param. Only the URL's `?project=` (always numeric) is forwarded.

All 1,413 tests pass.

_This response was AI-generated and verified against the updated code and passing tests._

PR #1002, opened 2026-05-26
chore(mcp-adoption-value-discovery): Adding docs traffic sources attrs

Abdkhan14:
Enables Sentry browser tracing on the SPA so we get pageload transactions for mcp.sentry.dev/ traffic. Added two attrs for traffic source tracking:

`app.utm_source `— sanitized in-product utm_source query param from window.location. Used to group by in-app sources as we surface mcp doc links more in Sentry. 

`app.referrer.family `— low-cardinality bucket of the Referer header host: google, github, sentry, sentry-docs, or other for now. Used to see where external traffic is arriving from.

### DIFF
--- a/packages/mcp-cloudflare/src/client/instrument.ts
+++ b/packages/mcp-cloudflare/src/client/instrument.ts
@@ -1,15 +1,14 @@
 import * as Sentry from "@sentry/react";
 import { sentryBeforeSend } from "@sentry/mcp-core/telem/sentry";
-import { resolveAttribution } from "./utils";
+import { attributionBeforeSendSpan } from "./utils";
 
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
   sendDefaultPii: true,
   tracesSampleRate: 1,
   beforeSend: sentryBeforeSend,
+  beforeSendSpan: attributionBeforeSendSpan,
   environment:
     import.meta.env.VITE_SENTRY_ENVIRONMENT ?? import.meta.env.NODE_ENV,
   integrations: [Sentry.browserTracingIntegration()],
 });
-
-resolveAttribution();
--- a/packages/mcp-cloudflare/src/client/instrument.ts
+++ b/packages/mcp-cloudflare/src/client/instrument.ts
@@ -1,5 +1,6 @@
 import * as Sentry from "@sentry/react";
 import { sentryBeforeSend } from "@sentry/mcp-core/telem/sentry";
+import { resolveAttribution } from "./utils";
 
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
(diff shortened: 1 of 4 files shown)


packages/mcp-cloudflare/src/client/instrument.ts
```diff
@@ -8,4 +9,7 @@ Sentry.init({
   beforeSend: sentryBeforeSend,
   environment:
     import.meta.env.VITE_SENTRY_ENVIRONMENT ?? import.meta.env.NODE_ENV,
+  integrations: [Sentry.browserTracingIntegration()],
 });
+
+resolveAttribution();
```
jaydgoss, line 15: `beforeSendSpan` might be a good alternative here, removes the assumption that the pageload span is active when `resolveAttribution` runs.

Something like:

```typescript
const utmSource = resolveUtmSource(
  new URLSearchParams(window.location.search).get("utm_source"),
);
const referrerFamily = resolveReferrerFamily(document.referrer);

Sentry.init({
  // ...
  integrations: [Sentry.browserTracingIntegration()],
  beforeSendSpan(span) {
    if (span.op === "pageload") {
      span.data = {
        ...span.data,
        ...(utmSource && { "app.utm_source": utmSource }),
        ...(referrerFamily && { "app.referrer.family": referrerFamily }),
      };
    }
    return span;
  },
});
```

jaydgoss (approved):

packages/mcp-cloudflare/src/client/instrument.ts
Abdkhan14: @jaydgoss great callout, pushed the change thanks 💯

dcramer: fwiw im not a fan of beforeSendSpan and that kind of magic behavior. its way too brittle. much safer to just bind it explicitly where its needed so it doesnt randomly break in the future when e.g. op gets renamed or something like that

packages/mcp-cloudflare/src/client/instrument.ts
```diff
@@ -1,11 +1,14 @@
...
 
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
   sendDefaultPii: true,
   tracesSampleRate: 1,
   beforeSend: sentryBeforeSend,
+  beforeSendSpan: attributionBeforeSendSpan,
   environment:
     import.meta.env.VITE_SENTRY_ENVIRONMENT ?? import.meta.env.NODE_ENV,
+  integrations: [Sentry.browserTracingIntegration()],
```
dcramer, line 12: i think i had this disabled because the data is mostly inactionable/junk. we can see how it goes but i want to make sure the quality of the telemetry stays as high as it can on this project

packages/mcp-cloudflare/src/client/instrument.ts
Abdkhan14: yeah I understand ^, beforeSendSpan + op filter does end up coupled to internal SDK strings. 

Reverting to explicit getActiveSpan + setAttribute right after init in the next push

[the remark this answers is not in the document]
Abdkhan14: We rely on browserTracingIntegration starting the pageload span synchronously after init

PR #1013, opened 2026-05-27
feat: Add repository linking to create_project tool

wedamija:
- Add RepositorySchema, RepositoryListSchema, ProjectRepoLinkSchema
- Add listRepos() and linkProjectRepo() to API client
- Add optional `repository` param to create_project tool
- When provided, searches for the repo by name and links it to the newly created project via POST /projects/{org}/{project}/repo/


packages/mcp-core/src/api-client/schema.ts
```diff
@@ -123,6 +123,35 @@ export const ProjectSchema = z
...
+        id: z.string(),
+        name: z.string(),
+      })
+      .passthrough(),
+    status: z.string(),
+    externalSlug: z.string().optional(),
+    externalId: z.string().optional(),
+    integrationId: z.union([z.string(), z.number()]).nullable().optional(),
+  })
+  .passthrough();
```
evanpurkhiser, line 141: Too bad we can't get our API story together and generate these from the OpenAPI schema we export from sentry.

packages/mcp-core/src/tools/create-project.ts
```diff
@@ -34,11 +35,16 @@ export default defineTool({
...
+    "### Create project and link to a repository",
+    "```",
+    "create_project(organizationSlug='my-organization', teamSlug='my-team', name='my-project', platform='javascript', repository='getsentry/sentry')",
+    "```",
     "</examples>",
     "",
     "<hints>",
     "- If the user passes a parameter in the form of name/otherName, its likely in the format of <organizationSlug>/<teamSlug>.",
     "- If any parameter is ambiguous, you should clarify with the user what they meant.",
+    "- The repository parameter accepts a repo name (e.g. 'getsentry/sentry'). The repo must already be connected to the org via a VCS integration.",
```
evanpurkhiser, line 47: We don't have a tool to do this yet though right. The LLM would need to run list repos first to see what repos are available.

dcramer (approved):

packages/mcp-core/src/tools/create-project.ts
```diff
@@ -34,11 +35,16 @@ export default defineTool({
     "```",
     "create_project(organizationSlug='my-organization', teamSlug='my-team', name='my-project', platform='javascript')",
     "```",
+    "### Create project and link to a repository",
+    "```",
+    "create_project(organizationSlug='my-organization', teamSlug='my-team', name='my-project', platform='javascript', repository='getsentry/sentry')",
+    "```",
```
evanpurkhiser, line 41: I think we should give the LLM more context here. When creating a project it should probably by default to trying to figure out if it's in a git repository and then should be given some examples of how to infer the sentry repository name, eg:

```
git@github.com:orgSlug/repoName → orgSlug/repoName
gitlab.com/orgSlug/RepoName → orgSlug/repoName
bitBucket → (I think this one is different we need to check)
```

Failures should probably indicate something like:

> use the list repositories tool to see what VCS repositories are linked to the sentry organization, if the repository is not available the user will need to install the required Integration

dcramer: thats probably too much steering. it has a lot of that in its corpus and this is a single thing.

e.g. we dont want to use half our token budget on describing repo names

evanpurkhiser: Yeah sounds good, I think we should try to get it to include the repository by default though

packages/mcp-core/src/skillDefinitions.json
```diff
@@ -224,7 +224,7 @@
...
         "requiredScopes": ["project:write", "team:read"]
       },
```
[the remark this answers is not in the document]
dcramer: good catch warden - i think this is valid

PR #1017, opened 2026-05-28
fix(search-events): Drop dependency on private validate endpoint

skaasten:
## Summary

- `datasetAttributes` (the embedded agent tool inside `search_events`) was calling `POST .../trace-items/attributes/validate/`, an `ApiPublishStatus.PRIVATE` Sentry endpoint whose default `scope_map` requires `org:write` for POST — a scope MCP's OAuth tokens never request.
- The result was a silent 403 on every `search_events` call that touched the validation path. The throw collapsed the tool's `execute()` before `fetchCustomAttributes` could run, leaving the agent with no field discovery for that turn.
- Switches verification to the public `GET .../trace-items/attributes/` path via `substringMatch`, which the tool already uses for broader field discovery. `org:read` only, no PRIVATE-endpoint dependency.

## What changed

- `datasetAttributes` tool: drops the `attributes` exact-validation parameter; agent now verifies a specific field by calling the same tool with `substringMatch=<exact-name>`.
- Agent prompts in `config.ts` and the repair prompt in `handler.ts` updated accordingly.
- Removed `validateTraceItemAttributes` from the API client, its unit test, and the MSW mock.
- Added a regression test asserting the validate endpoint is never called.

## Test plan

- [ ] `pnpm run tsc && pnpm run lint && pnpm run test` passes (green locally)
- [ ] Manual `search_events` query against an OAuth-scoped MCP token returns results without 403 spans on validate
- [ ] Spot-check a `tags[...]` lookup end-to-end to confirm custom attributes still resolve


dcramer (approved):
i didnt verify the actual endpoint, but broadly speaking any change that improves the query capabilities is +1

DominikB2014 (approved):

PR #1025, opened 2026-05-29
perf(search-events): Skip attribute API lookup for built-in fields

skaasten:
## Summary

`datasetAttributes` (the embedded-agent tool inside `search_events`) verified every explicit field by calling the trace-items attributes endpoint — even for **built-in fields** like `span.duration` or `http.method` that are already enumerated in static config. The prompt encourages the agent to confirm each explicit field via `substringMatch`, so a single agent turn could stack up a dozen-plus sequential attribute lookups.

That redundant network fan-out lengthened discovery-heavy turns enough to occasionally tip a `search_events` call into an `AI_NoOutputGeneratedError` (the agent exhausting its time/step budget before producing output).

## What changed

- **Tool short-circuit (`utils.ts`):** build a `Set` of known built-in field names (`BASE_COMMON_FIELDS` ∪ `DATASET_FIELDS[dataset]` ∪ `NUMERIC_FIELDS[dataset]`). If `substringMatch` exactly matches a built-in, skip `fetchCustomAttributes` entirely — the field is already confirmed from config, no GET to `/trace-items/attributes/`. Own-keys-only `Set.has` is used (the match target is untrusted input, so prototype-chain checks like `in` are avoided).
- The output's Custom Attributes section now reports the skip explicitly instead of rendering a misleading empty result.
- **Prompt guidance (`config.ts`):** tell the agent that fields in the "Built-in Fields" listing are pre-confirmed, and reserve `substringMatch` for suspected custom attributes (`tags[...]`) not already in that list.

## Effect

Built-in field checks now resolve in-memory with zero network round-trips; only genuine custom-attribute lookups still hit the API. This is the path that no-ops the validate-endpoint dependency removed in #1017 — same field-discovery flow, far fewer calls.

## Test plan

- [x] New regression test: verifying a built-in field (`span.duration`) makes **no** call to the attributes endpoint and reports it as built-in.
- [x] Existing search-events tests still pass (38), incl. the custom-attribute (`tags[...]`) discovery and static-field-cap guards.
- [x] `pnpm run tsc` and `pnpm run lint` clean.

### DIFF
--- a/packages/mcp-core/src/tools/search-events/config.ts
+++ b/packages/mcp-core/src/tools/search-events/config.ts
@@ -28,7 +28,7 @@ Before constructing ANY query, you MUST verify field availability:
 2. This includes ALL fields: custom attributes, database fields, HTTP fields, AI fields, user fields, etc.
 3. Fields vary by project based on what data is being sent to Sentry
 4. Using an unverified field WILL cause your query to fail with "field not found" errors
-5. For spans, logs, and metrics, datasetAttributes lists fields and supports substringMatch/query/attributeTypes for targeted lookup; pass the exact field name as substringMatch to confirm whether it exists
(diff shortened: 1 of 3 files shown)


DominikB2014 (commented):
The attribute endpoint also validates the data is there which is pretty valuable imo, and additionallythere's some edge cases to consider where users send in custom attributes similar to sentry built ones. I think bulk validation might be the better fix. Lets chat about this and see what makes sense

PR #1037, opened 2026-06-01
ci(deploy): wire VITE_SENTRY_DSN into the build env

Abdkhan14:
The SPA's Sentry.init() reads `import.meta.env.VITE_SENTRY_DSN`, which Vite inlines at build time. The deploy workflow never exported this variable, so every production bundle has shipped with `dsn: undefined` (minified to `void 0` in the bundle). Suspecting this is silently breaking SPA telemetry 

This wires `secrets.VITE_SENTRY_DSN` and a hardcoded `VITE_SENTRY_ENVIRONMENT=production` into the Build step. Both keys are already declared in turbo.json's globalPassThroughEnv and documented in TELEMETRY.md as required build-time vars.

I have already assigned VITE_SENTRY_DSN: {dsn from project settings} as a secret to the sentry-mcp repo using gh cli

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -54,6 +54,8 @@ jobs:
         run: pnpm -w run build
         env:
           SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          VITE_SENTRY_DSN: ${{ secrets.VITE_SENTRY_DSN }}
+          VITE_SENTRY_ENVIRONMENT: production
 
       - name: Deploy to Canary Worker
         id: deploy_canary


dcramer:
we bind VITE_SENTRY_DSN in cloudflare already

this would mean we have to maintain it in two places

Abdkhan14:
> we bind VITE_SENTRY_DSN in cloudflare already
> 
> this would mean we have to maintain it in two places

@dcramer Good callout on duplication. Just putting out my understanding for validation: these are two different stages, the worker reads env.SENTRY_DSN at runtime (that's the Cloudflare binding, and it works), but the SPA's Sentry.init() reads import.meta.env.VITE_SENTRY_DSN, which Vite inlines at build time. The build runs in GH Actions, which can't see Cloudflare's bindings — so the value we bind in Cloudflare never reaches the browser bundle. That's why pageload telemetry for mcp.sentry.dev has been silently dead while worker telemetry is fine.

So our options are (a) expose the DSN to the build, which is this PR but does mean a second place, or (b) make Cloudflare the single source of truth by serving the DSN to the SPA at runtime instead of inlining it — the worker already has it in env, so it can inject it into index.html (or maybe through an endpoint) and the client reads it at runtime. 

Happy to switch this PR to (b) if you prefer the single-source approach.

dcramer:
Ah ok we may want to find another way to clean all this up but i dont have an obvious better solution for now then

PR #1040, opened 2026-06-02
feat: Improve agent guidance for AI conversation queries

obostjancic:
Adds conversation concept guidance to tool descriptions and plugin agent prompts so agents route conversation queries to `get_sentry_resource` and `search_events` instead of misusing `search_issues` or using wrong attributes like `span.op`.

It would be good to add something also to `search_events` description but its description is 11 chars under the 2048 limit so I skipped that for now.

Closes TET-2414


dcramer:
going to f/u later with an attempt to fix the fact that we cant fi teverything into these at this point

PR #1043, opened 2026-06-04
feat(mcp): Collapse preprod into inspect experimentally

dcramer:
Experimental MCP sessions now expose preprod snapshot tools through the existing inspect permission instead of requiring a separate preprod grant. Stable sessions keep preprod as its own permission.

**Consent and Scopes**

Experimental OAuth resource URLs hide the separate Preprod Snapshots option and compute granted scopes with the same mode-aware skill merge used by tool availability.

**Catalog Availability**

The catalog runtime expands merged skills only in experimental mode, so preprod tools can be discovered through search_tools while remaining off the default top-level tool list.


dcramer:
Closing as completed by later work. PR #1067 promoted the catalog gateway and consolidated the skill behavior, including compatibility mapping from the former `preprod` grant to `inspect`.

PR #1045, opened 2026-06-04
feat(mcp): Add structured Sentry tool results

dcramer:
Add experimental structuredContent result payloads for Sentry issue details and event search while preserving serialized JSON text fallback. This lets clients consume typed Sentry data without parsing Markdown, while older content-only adapters continue to receive semantically equivalent results.

**Structured Result Contract**

Move `get_issue_details`, `search_events`, and `search_issue_events` onto schema-versioned structured result payloads in experimental mode. The shared tool-result path keeps MCP `content` equivalent to `structuredContent`, and catalog discovery now exposes output schemas for structured targets.

**Untrusted Telemetry Handling**

Use one broad security note for structured Sentry payloads instead of field-level unsafe path lists. Issue-event structured rows now reuse the Markdown renderer row contract rather than returning raw event API objects, so user-controlled telemetry remains formatted result data instead of becoming a larger raw prompt-injection surface.

**Regression Coverage And Audit Guidance**

Add structured snapshots and prompt-injection canaries around issue details, plus targeted coverage for issue-event structured rows with instruction-like telemetry. Update the MCP audit skill and common patterns docs so future structuredContent migrations check untrusted data handling and preserve the rendered-result contract.


dcramer:
Closing as superseded. Structured results are now being migrated incrementally per tool through the #1193 rollout and merged PRs such as #1195, #1196, and #1198-#1204. Reviving this broad experimental branch would conflict with the current pattern.

PR #1053, opened 2026-06-05
Revert "feat(seer): Add referrer to autofix start request"

chromy:
This reverts commit e22f4094329ca2ebf5410d8a84329ee0a58efd64.


dcramer:
context: id advise we just look at which applications are making calls vs hardcoding things (esp given the ecosystem isnt going to send referer values for us)

most mcp traffic comes from a single oauth app (mcp.sentry.dev)

dcramer (approved):

Issue #1064, opened 2026-06-08
analyze_issue_with_seer fails with ZodError: Failed to validate keys: autofix.status


InterstellarStella:
## What happened

Calling `analyze_issue_with_seer` on a SaaS org consistently fails. The agent surfaces a generic "Sentry API communication error", but the underlying MCP telemetry shows the real error:

```
ZodError: Failed to validate keys: autofix.status
  at SentryApiService.getAutofixState
```

Reproduces on both transports (cloud `mcp.sentry.dev` and stdio `@sentry/mcp-server@latest`). Read-only tools and the Sentry UI's Seer work fine.

## Root cause

`GET /organizations/{org}/issues/{id}/autofix/` now returns the explorer/agent-based shape with a **lowercase** status (`processing` / `completed` / `error` / `awaiting_user_input`), but `AutofixStatusSchema` only accepts the legacy **uppercase** enum (`PROCESSING`, `COMPLETED`, …), so `AutofixRunStateSchema.parse()` throws. Since `ZodError` isn't an `ApiError`, it falls through to the generic "communication error" branch instead of a clear validation message. Not a 5xx, and unrelated to billing/quota (that path returns 402).

## Fix already exists but is unmerged

This is exactly what the closed #961 addressed, now split into the draft stack #963 + #966 ("Status enum collapsed to the explorer values"). Both have been sitting in draft since 2026-05-13. Filing this to track landing + releasing them, since the published package and cloud deployment are currently broken for any org on the explorer-mode autofix backend.

PR #1070, opened 2026-06-09
fix(seer): use updated autofix status values

cvxluo:
getsentry/sentry#116162 updated `GroupAutofixEndpoint` to use the new explorer-based autofix, which has a new schema. When the MCP attempted to use seer tools, they would always fail validation. Update to use the new autofix schema.

Fixes MCP-SERVER-FXM


dcramer:
fixes https://github.com/getsentry/sentry-mcp/issues/1064

PR #1101, opened 2026-06-17
fix(schema): accept relative permalinks from self-hosted Sentry

n8tron450:
## Problem
`IssueSchema.permalink` is validated with `z.string().url()`. Self-hosted Sentry returns **relative** permalinks (e.g. `/organizations/<org>/issues/<id>/`) rather than absolute URLs, so validation throws `Invalid url` on the `permalink` path. This makes `search_issues` (and any tool returning issues) fail entirely against self-hosted instances (e.g. `--host localhost:9000 --insecure-http`).

## Fix
Relax `permalink` to `z.string()`. The field is display-only and never parsed as a URL, so URL validation adds no value while breaking self-hosted users.

## Testing
Verified against a self-hosted Sentry instance: `search_issues` returns results instead of erroring on permalink validation.

### DIFF
--- a/packages/mcp-core/src/api-client/schema.ts
+++ b/packages/mcp-core/src/api-client/schema.ts
@@ -702,7 +702,7 @@ export const IssueSchema = z
     lastSeen: z.string().datetime().nullable(),
     count: z.union([z.string(), z.number()]),
     userCount: z.union([z.string(), z.number()]),
-    permalink: z.string().url(),
+    permalink: z.string(), // relative on self-hosted Sentry; .url() rejects it (permalink is display-only)
     project: ProjectSchema,
     platform: z.string().nullable().optional(),
     status: z.string(),


dcramer:
Traced through the Sentry source to verify — the bug is real, but there are actually two mechanisms at play:

**Path 1 — unconfigured `system.url-prefix` (likely for dev/testing setups)**
- `Group.get_absolute_url()` → `organization.absolute_url()` → `absolute_uri(path, url_prefix=...)`
- `absolute_uri` falls back to `options.get("system.url-prefix")`, which defaults to `os.environ.get("SENTRY_SYSTEM_URL_PREFIX")`
- if that env var isn't set, `url_prefix` is empty, and `urljoin("/", "organizations/myorg/issues/123/")` produces the literal relative path `/organizations/myorg/issues/123/`
- this is where the "relative URL" claim in the PR description is accurate

**Path 2 — `system.url-prefix` is set to `http://localhost:9000`**
- the result IS an absolute URL (`http://localhost:9000/organizations/...`)
- but Zod v3's `.url()` validator uses a regex that requires a dotted hostname (e.g. `sentry.io`) — `localhost` fails validation
- this is a known Zod footgun

The `--host localhost:9000 --insecure-http` test setup would hit Path 2 (or both).

**Fix is sound:** `permalink` is display-only — it's returned in the serialized output and never parsed as a URL in MCP. Relaxing to `z.string()` is the right call.

dcramer:
the above is one of our agents, but fwiw you need to configure the url prefix on a self hosted install or a bunch of other things will break (like outbound emails)

i think id rather let this break because otherwise we have all sorts of edge cases we have to handle here that should be handled upstream

dcramer:
going to close as a wontfix but happy to hear arguments otherwise

PR #1118, opened 2026-06-22
feat(api): Add MCP referrer to search_events API calls

wmak:
Set referrer=api.mcp.search-events on Discover and EAP /events/ requests so Sentry can attribute search_events traffic.


dcramer:
at some point we should figure out how to avoid doing this, since our customers wont do it for us, and we already have api application attribution capabililty on the backend

wmak:
Makes sense, I'll revisit this later to see if i can improve it

PR #1120, opened 2026-06-22
build: Bump Sentry JS SDK to 10.61.0 and enable span streaming

cleptric:
We're dropping a small number of transactions/spans due to size limits. Migrating to span streaming to improve this.

---

Upgrade `@sentry/cloudflare`, `@sentry/core`, `@sentry/node`, and `@sentry/react`
from 10.54.0 to 10.61.0 and enable the span streaming trace lifecycle so spans
are sent in batches as they complete instead of being emitted when the transaction finishes.

Server-side configs (`mcp-server`, `mcp-test-client`, cloudflare worker) get
`traceLifecycle: 'stream'`. The browser client (`mcp-cloudflare`) gets
`spanStreamingIntegration()` added to the integrations array, which enables
streaming automatically.

### DIFF
--- a/packages/mcp-cloudflare/src/server/sentry.config.ts
+++ b/packages/mcp-cloudflare/src/server/sentry.config.ts
@@ -10,6 +10,7 @@ export default function getSentryConfig(env: Env): CloudflareOptions {
   return {
     dsn: env.SENTRY_DSN,
     tracesSampleRate: 1,
+    traceLifecycle: "stream",
     sendDefaultPii: true,
     beforeSend: sentryBeforeSend,
     initialScope: {
--- a/packages/mcp-cloudflare/src/server/sentry.config.ts
+++ b/packages/mcp-cloudflare/src/server/sentry.config.ts
@@ -18,6 +18,10 @@ export default function getSentryConfig(env: Env): CloudflareOptions {
         "app.server.version": LIB_VERSION,
         "app.upstream.host": env.SENTRY_HOST,
       },
+      attributes: {
+        "app.server.version": LIB_VERSION,
+        "app.upstream.host": env.SENTRY_HOST,
+      },
     },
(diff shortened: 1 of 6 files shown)


packages/mcp-cloudflare/src/server/sentry.config.ts
```diff
@@ -10,6 +10,7 @@ export default function getSentryConfig(env: Env): CloudflareOptions {
   return {
     dsn: env.SENTRY_DSN,
     tracesSampleRate: 1,
+    traceLifecycle: "stream",
```
Lms24, line 13: l: below we can mirror the `initialScope.tags` to `initialScope.attributes`

pnpm-workspace.yaml
```diff
@@ -35,10 +35,10 @@ catalog:
...
   "@radix-ui/react-accordion": ^1.2.11
   "@radix-ui/react-slot": ^1.2.3
-  "@sentry/cloudflare": 10.54.0
-  "@sentry/core": 10.54.0
-  "@sentry/node": 10.54.0
-  "@sentry/react": 10.54.0
+  "@sentry/cloudflare": 10.59.0
+  "@sentry/core": 10.59.0
+  "@sentry/node": 10.59.0
+  "@sentry/react": 10.59.0
```
Lms24: I'd recommend going straight to 10.61.0 for `Sentry.setAttribute`

Lms24 (approved):

Lms24 (approved):
It seems like we call `setTag` a bunch of times. we likely want to convert these calls to attributes. The [skill](https://github.com/getsentry/sentry-for-ai/blob/main/skills/sentry-span-streaming-js/SKILL.md) was already updated to make this migration but since our release to agent plugin marketplaces is broken, you might need to run the skill manually.

Besides the setAttribute calls, LGTM, so ✅ing.

PR #1121, opened 2026-06-23
feat(search-events): Query available fields before constructing a query

wmak:
- Queries aren't fetching fields first before constructing a query which is leading to a decent number of the invalid queries beind sent to /validate
- Query available fields first to see if this improves valid queries


wmak:
hmm this seems to be the wrong approach closing for now

PR #1126, opened 2026-06-26
fix: surface user identity in AI conversations and default error search fields

### LINKED ISSUE #1033: get_ai_conversation_details drops user.email from transcript; search_events omits user identity from default error fields
## Problem

When using the MCP to debug AI agent runs (e.g. Seer), you can't tie failures back to specific users or prompts. Josh Ferge ran into this directly: *"Sentry MCP dropped `user_email` / `user_message` from results, so I couldn't tie specific failures to user prompts."*

There are two independent bugs causing this, both traceable to specific lines of code.

---

### Root cause

`extractTurns()` correctly reads `user.email` from each span and stores it in `ConversationMessage.userEmail`:

```typescript
// get-ai-conversation-details.ts ~line 261
user: userContent
  ? {
      role: "user",
      content: userContent,
      timestamp: span["precise.start_ts"],
      spanId: span.span_id,
      userEmail: span["user.email"],  // ← collected ✓
    }
  : undefined,
```

But `formatTurn()` never renders it:

```typescript
// get-ai-conversation-details.ts ~lines 349-352
turn.user ? "**User**" : null,
turn.user ? "" : null,
turn.user ? truncate(turn.user.content) : null,  // content only
turn.user ? "" : null,
// userEmail is silently dropped ✗
```

The email IS present in the raw `artifact` JSON blob appended at the end of the response — but that's buried in a ~1200-char code block agents have to parse to find a single field. The transcript, which is what agents actually read turn-by-turn, has no user identity.

### Impact

An agent debugging Seer runs can see the full conversation transcript but has no way to answer "which user triggered this run?" without separately parsing the JSON artifact blob at the bottom.

### Fix

One-line change in `formatTurn()`:

```typescript
turn.user ? `**User**${turn.user.userEmail ? ` (${turn.user.userEmail})` : ""}` : null,
```

Plus snapshot test update.

---

### Root cause

The recommended field set for error queries (`RECOMMENDED_FIELDS.errors.basic` in `search-events/config.ts`) is:

```typescript
errors: {
  basic: ["issue", "title", "project", "timestamp", "level", "message", "error.type", "culprit"],
}
```

No user fields. The search sub-agent's default error query returns events with zero user identity context.

`user.email` and `user.id` ARE in `BASE_COMMON_FIELDS` (surfaced to the sub-agent via `datasetAttributes`), so the agent _can_ request them if the user explicitly asks. But recommended fields are the fallback default — and they omit user identity entirely.

Net result: "show me recent Seer errors" returns a list of failures with no indication of who triggered them.

### Fix

Add user fields to the recommended set:

```typescript
errors: {
  basic: [
    "issue", "title", "project", "timestamp", "level",
    "message", "error.type", "culprit",
    "user.email",  // add
    "user.id",     // add
  ],
}
```

---

## What's Out of Scope Here

The deeper gap is that **Seer Celery task error events** may not carry user context at all. If the Seer task doesn't attach `user.email` and the original prompt to the Sentry error event when it crashes or gets SIGKILL'd, no MCP fix can recover that data. That's a Seer instrumentation gap in `getsentry/sentry`: the Seer task execution scope should set Sentry user context and include the originating user message as a tag/extra, so error events carry it.

The two fixes above address what the MCP can control: surfacing data that _is_ present in spans, and requesting it by default.

---

## Acceptance Criteria

- [ ] `get_ai_conversation_details` transcript shows `user.email` inline on the **User** line (e.g. `**User** (alice@acme.com)`) when the field is set on the span
- [ ] `RECOMMENDED_FIELDS.errors.basic` includes `user.email` and `user.id`
- [ ] Snapshot tests updated
- [ ] Manual test: querying errors returns user email without needing to specify the field explicitly

---

## References

- Discussion: https://sentry.slack.com/archives/C089L9UUJDV/p1779997855795739
- Josh Ferge's Seer debug run: https://sentry.slack.com/archives/C091QF41ULT/p1778159743300829
- Related (prompt injection warnings): https://github.com/getsentry/sentry-mcp/issues/1019

---

[View Session in Sentry](https://sentry.sentry.io/traces/?project=4510944073809921&query=gen_ai.conversation.id%3A%22slack%3AC089L9UUJDV%3A1779997855.795739%22)

syf2211:
## Summary

Expose `user.email` on user timeline messages in `get_ai_conversation_details`, and include `user.email` / `user.id` in the default `search_events` error field set.

## Motivation

When debugging AI agent runs (e.g. Seer), maintainers could not tie failures back to specific users because:

1. `get_ai_conversation_details` collected `user.email` from spans but did not project it into the timeline messages agents actually read.
2. `search_events` default error fields omitted user identity, so error listings lacked user context unless explicitly requested.

Fixes #1033

## Changes

- Add optional `userEmail` to conversation timeline user messages (from `span["user.email"]`, omitted when absent).
- Extend `RECOMMENDED_FIELDS.errors.basic` with `user.email` and `user.id`.
- Regenerate `toolDefinitions.json` for the updated output schema.
- Add regression tests and snapshot update.

## Tests

```bash
pnpm --filter @sentry/mcp-core test -- src/tools/catalog/get-ai-conversation-details.test.ts src/tools/catalog/search-events.test.ts
```

Result: **1212 passed** (mcp-core suite)

## Notes

- `userEmail` is only included when present on the span; no placeholder values.
- This restores debugging context already available in Sentry spans under existing `event:read` scopes.

### DIFF
--- a/docs/specs/ai-conversations.md
+++ b/docs/specs/ai-conversations.md
@@ -223,6 +223,8 @@ Current expected behavior:
   default `statsPeriod` or explicit `start` and `end` values.
 - User and assistant message events are extracted from GenAI input/output
   fields.
+- User message events include `userEmail` when the source span has
+  `user.email`.
 - Tool call events are represented as first-class timeline events, not nested
   under a synthetic turn.
 - Each timeline event includes enough trace context to debug it, including
--- a/docs/specs/search-events.md
+++ b/docs/specs/search-events.md
@@ -69,6 +69,7 @@ search_events({
 4. **Executes** discover endpoint: `/organizations/{org}/events/` with:
    - Translated query string
    - Dataset-specific field selection
+   - Default error field selection includes `user.email` and `user.id`
    - Numeric project ID (converted from slug if provided)
    - Public dataset normalization (`metrics` maps to the current API dataset `tracemetrics`)
 5. **Returns** formatted results with:
--- a/packages/mcp-core/src/toolDefinitions.json
+++ b/packages/mcp-core/src/toolDefinitions.json
@@ -837,6 +837,9 @@
                   "traceId": {
                     "type": "string"
                   },
+                  "userEmail": {
+                    "type": "string"
+                  },
                   "metadata": {
                     "type": "object",
                     "properties": {
--- a/packages/mcp-core/src/tools/catalog/get-ai-conversation-details.test.ts
+++ b/packages/mcp-core/src/tools/catalog/get-ai-conversation-details.test.ts
@@ -140,6 +140,7 @@ describe("get_ai_conversation_details", () => {
             "timestamp": 1713805400000,
             "traceId": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
             "type": "message",
+            "userEmail": "dev@example.com",
           },
           {
             "content": "The checkout worker is timing out.",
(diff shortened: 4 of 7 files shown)


dcramer:
thanks!

PR #1129, opened 2026-06-26
fix(formatting): include metric alert rule details in issue output

### LINKED ISSUE #1087: Metric issue output missing alert rule details (query, aggregate, condition)
Metric issue tool output omits alert rule details (query, aggregate, condition, evaluated value), leaving the agent without the context it needs to help root-cause an alert.

The Sentry UI surfaces this clearly (see example: dataset, aggregate `count()`, query `process_shard OR OutboxFlushError`, interval, threshold, evaluated value), but none of it is included in the MCP tool response for `metric_issue` type issues.

- Issue type: `metric_issue` / category: `metric`
- Missing fields: alert rule query, dataset, aggregate function, interval, threshold condition, evaluated value
- Impact: agent responses are effectively useless for metric alerts — no query context means no ability to investigate signal vs. noise, trend, or root cause
- The event output currently includes only generic tags (`level`, `interface_type`) and trace info, with no alert rule payload

syf2211:
## Summary

Surface metric alert rule context in `get_issue_details` output for `metric_issue` events. Agents now see dataset, aggregate, query, interval, environment, evaluated value, threshold, and alert rule ID parsed from occurrence `evidenceData`.

## Motivation

Fixes #1087. Metric alert issues previously returned only generic tags and trace info, leaving agents without the query/aggregate/threshold context shown in the Sentry UI.

## Changes

- Detect metric alert evidence in `formatGenericEventOutput` via `data_sources[].query_obj.snuba_query` or `alert_id`
- Add `### Metric Alert Details` section with alert rule fields from workflow-engine evidence payloads
- Map workflow-engine threshold types (`gt`/`gte`/`lt`/`lte`, plus legacy numeric types) to readable threshold text
- Keep existing performance regression formatting unchanged

## Tests

- `pnpm exec vitest run src/internal/formatting.test.ts` — 53 passed
- `pnpm exec vitest run src/tools/catalog/get-issue-details.test.ts` — 22 passed (3 skipped)
- `pnpm exec biome check` on changed files — clean

## Notes

- Formatting-only change; no new API calls or auth paths
- When only `alert_id` is present (no `data_sources`), evaluated value/threshold still render; query fields are omitted


dcramer:
At some point we have to fix these magic numbers from the API, but I will f/u internally on that.

PR #1141, opened 2026-07-02
feat(catalog): add get_issue_user_reports tool

Pragyan14:
## What

Exposes `GET /api/0/issues/{issue_id}/user-reports/` through the MCP catalog as a new `get_issue_user_reports` tool.

## Changes

- `schema.ts` — adds `UserReportSchema` and `UserReportListSchema` matching the real API response shape
- `client.ts` — adds `getIssueUserReports()` method using the org-scoped endpoint `/organizations/{org}/issues/{id}/user-reports/`
- `types.ts` — adds `UserReportList` type inferred from the list schema
- `tools/catalog/get-issue-user-reports.ts` — new catalog tool following existing patterns (`get-issue-activity`, `get-issue-tag-values`)
- `tools/catalog/index.ts` — registers the new tool
- `mcp-server-mocks` — adds mock handler + `issue-user-reports.json` fixture based on real API response

## Testing

- Reproduced the gap against a real sentry.io project: confirmed `GET /issues/{id}/user-reports/` returns feedback data that was completely invisible to agents before this change
- Confirmed org-scoped URL (`/organizations/{org}/issues/{id}/user-reports/`) returns identical data
- Unit test passes with snapshot matching real formatted output
- `tsc` and `lint` pass clean on `@sentry/mcp-core`
- Tool shows as accessible in catalog summary (43 tools, 0 orphaned)

## Notes

- Tool is exposed via catalog only (behind `search_sentry_tools` / `execute_sentry_tool`), per maintainer guidance
- One pre-existing test failure in `@sentry/mcp-agent-cli-test` (`setup.test.ts`) is a Windows path normalization bug unrelated to this change — confirmed by reproducing with a clean stash

### DIFF
--- a/packages/mcp-core/src/api-client/client.ts
+++ b/packages/mcp-core/src/api-client/client.ts
@@ -79,6 +79,7 @@ import {
   ReplayRecordingSegmentsSchema,
   AIConversationSummaryListSchema,
   AIConversationSpanListSchema,
+  UserReportListSchema,
 } from "./schema";
 import { ConfigurationError } from "../errors";
 import { createApiError, ApiNotFoundError, ApiValidationError } from "./errors";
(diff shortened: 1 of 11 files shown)


dcramer:
cleaning up a couploe things. one thing i generally do is point the agent at current sentry to make sure theres no gaps in schemas, but broadly speaking patch is good and we'll get it landed today, ty!

PR #1144, opened 2026-07-02
Enforce annotations for readOnlyHint and destructiveHint

### LINKED ISSUE #1143: Enforce complete tool hint annotations with a linter
Every tool in the catalog must declare `readOnlyHint` and `destructiveHint` in its `annotations` block. Right now several write tools (`add-issue-note`, `analyze-issue-with-seer`, `create-dsn`, `create-project`, `create-team`) have no annotations at all, while others declare only one of the two hints. This creates ambiguity for any routing logic, client-side confirmation gates, or server-side filters that rely on these values.

**Scope**

- Audit all tools in `packages/mcp-core/src/tools/catalog/` and `catalog-runtime/` for missing or partial `annotations` blocks.
- Add `readOnlyHint` and `destructiveHint` to every tool that lacks them.
- Add a deterministic test or CI lint rule that fails if any registered tool is missing either hint — so gaps can't be reintroduced silently. Something like iterating `allTools` and asserting both fields are explicitly `true` or `false`, not `undefined`.

**Known gaps (as of reviewed commit `3705c15`)**

- `add-issue-note` — no `destructiveHint`
- `analyze-issue-with-seer` — no `readOnlyHint` or `destructiveHint`
- `create-dsn` — no `destructiveHint`
- `create-project` — no `destructiveHint`
- `create-team` — no `destructiveHint`
- `update-dsn`, `update-issue`, `update-project` — have `destructiveHint: true` but no explicit `readOnlyHint: false`

Related: `sentry-geoffg/readOnlyHint-changes` branch adds server-side filtering keyed on `readOnlyHint`; that filter is only as good as the annotations it reads.

--

[View Junior Session in Sentry](https://sentry.sentry.io/explore/conversations/slack%3AC08J1NSPU6S%3A1783020980.100799/?project=4510944073809921)

geoffg-sentry:
Fixes #1143. Every catalog tool now declares complete MCP safety annotations, and this is enforced so gaps can't be reintroduced.
- Require complete hints at the type level: `readOnlyHint`, `destructiveHint`, and `openWorldHint` are now required on `ToolConfig.annotations` (each must be an explicit `true`/`false`). A missing hint is now a `tsc` error.
- Backfill read-only tools: added the previously-omitted `destructiveHint: false` to every read-only tool (all `find_*`/`get_*`/`search_*`/`whoami`, plus `search_sentry_tools`). The write tools already declared both hints on `main`; the gaps listed against `3705c15` were already fixed.
- Strengthen the deterministic test: `tools.test.ts` now asserts all three hints are explicit booleans on every registered tool (not just when `readOnlyHint === false`) and that no tool is both read-only and destructive.
- Docs: updated `docs/contributing/adding-tools.md`, whose examples were teaching the incomplete pattern, to reflect the requirement.

Will follow up afterwards to deprecate use_sentry since tool catalogs are our preferred way now.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@ Sentry MCP is a Model Context Protocol server that exposes Sentry's error tracki
 ## Principles
 
 - **Type Safety**: Prefer strict types over `any` - they catch bugs and improve tooling. Use `unknown` for truly unknown types.
-- **Security**: Never log secrets. Validate external input. See docs/operations/security.md.
+- **Security**: Never log secrets. Validate external input. Treat tool output as untrusted; agentic surfaces (embedded agent, `/chat`) are read-only by default and writes require a trusted-caller opt-in. Every tool must declare complete safety annotations (`readOnlyHint`/`destructiveHint`/`openWorldHint`). See docs/operations/security.md.
 - **Simplicity**: Follow existing patterns. Check neighboring files before inventing new approaches.
 
 ## Constraints
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@ Sentry MCP is a Model Context Protocol server that exposes Sentry's error tracki
 ## Principles
 
 - **Type Safety**: Prefer strict types over `any` - they catch bugs and improve tooling. Use `unknown` for truly unknown types.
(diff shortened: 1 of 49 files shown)


dcramer:
so my only concern is the change to use_sentry

thats agent mode and it wraps all tools - of which users explicitly select what tools they want. so us not allowing write tools there would actually be counter to the intent users select

i think we should just make use_sentry always dangerous? to be fair, agent mode probably is gonna die as well. so we could consider just removing it at this point instead

geoffg-sentry:
Yeah that makes sense, good justification on why we'd want move in that direction too. I'm going to drop user_sentry from /chat keepAnyway so the demo is fully read-only instead,

geoffg-sentry:
And reverted that use_sentry change because we're just going to deprecate it afterward

PR #1161, opened 2026-07-16
feat(search-issues): Support recommended sort

roggenkemper:
Adds `recommended` (Sentry's prioritized ranking) to the `search_issues` sort options — the tool enum, the embedded search-issues agent, and the `listIssues` API client. The issues API already accepts `sort=recommended`; the agent tooling just didn't expose it, so an agent couldn't ask for Sentry's default prioritized ranking.

`recommended` needs no client-side flag: the endpoint accepts it unconditionally (mapping to the v2 scorer internally when the org has the experimental flag).

`toolDefinitions.json` is hand-synced to the schema change; CI `generate-definitions` will confirm it matches.

### DIFF
--- a/packages/mcp-core/src/api-client/client.ts
+++ b/packages/mcp-core/src/api-client/client.ts
@@ -3008,7 +3008,7 @@ export class SentryApiService {
    * @param params.organizationSlug Organization identifier
    * @param params.projectSlug Project identifier (optional, scopes to specific project)
    * @param params.query Sentry search query (e.g., "is:unresolved browser:chrome")
-   * @param params.sortBy Sort order ("user", "freq", "date", "new")
+   * @param params.sortBy Sort order ("user", "freq", "date", "new", "recommended")
    * @param opts Request options
    * @returns Array of issues with metadata and statistics
    *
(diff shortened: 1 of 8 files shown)


packages/mcp-core/src/api-client/client.ts
```diff
@@ -2826,7 +2826,7 @@ export class SentryApiService {
       organizationSlug: string;
       projectSlug?: string;
       query?: string | null;
-      sortBy?: "user" | "freq" | "date" | "new";
+      sortBy?: "user" | "freq" | "date" | "new" | "recommended";
```
dcramer, line 3045: **P2:** `recommended` will not use the feature-flagged v2 scorer when `projectSlug` is present. `listIssues()` switches to the deprecated project issues endpoint below, but Sentry's `recommended` → `recommended_v2` mapping exists only in `OrganizationGroupIndexEndpoint`; `ProjectGroupIndexEndpoint` calls `prep_search()` directly. This makes the same sort behave differently for project-scoped requests. Could we use the organization issues endpoint for project-scoped searches too and pass a project filter/ID?

dcramer (commented):
I found one project-scoping inconsistency worth addressing.

roggenkemper (commented):
Re-reviewed against the latest `main`: no additional findings. The earlier project-scoping concern is now addressed because `listIssues()` consistently uses the organization issues endpoint and passes `projectId` as a query parameter, so `recommended` reaches the endpoint that performs the org-level scorer selection. The schema, agent prompt/output, generated definitions, and request passthrough coverage are aligned.

The test/typecheck/security checks passed. The failing Warden rollup is infrastructure-only: `code-review` and `mcp-audit` stopped after repeated provider-unavailable errors and reported no findings.

PR #1166, opened 2026-07-18
fix(test-client): Make OAuth callback address configurable

### LINKED ISSUE #1139: OAuth callback hardcoded to localhost breaks authentication in VMs
The MCP test client's OAuth callback is hardcoded to `http://localhost:8765/callback`. When running inside a VM or container, the browser (often on the host) cannot reach this localhost address, so the authentication flow cannot complete.

**Findings**
- `packages/mcp-test-client/src/auth/oauth.ts` hardcodes `OAUTH_REDIRECT_URI = http://localhost:8765/callback` and starts the callback server on `127.0.0.1:8765`.
- There is no environment variable or CLI flag to override the redirect URI or callback port.
- Remote mode supports `SENTRY_ACCESS_TOKEN` as a workaround, but the OAuth flow itself is unusable in non-local environments without manual port forwarding.

JoshKappler:
Fixes #1139.

## Problem

The test client hardcodes the OAuth callback to `http://localhost:8765/callback` (`constants.ts`) and binds the callback server to `127.0.0.1` (`auth/oauth.ts`). Running inside a VM or container the browser lives on the host and cannot reach that address, and there is no env var or flag to override the port or the redirect URI, so the OAuth flow cannot complete. `SENTRY_ACCESS_TOKEN` only helps if you already have a token.

## Changes

Adds three variables, resolved once per `OAuthClient` in a new `auth/redirect.ts`:

| Variable | Default | Purpose |
| --- | --- | --- |
| `MCP_OAUTH_PORT` | `8765` | Port the callback server listens on |
| `MCP_OAUTH_HOST` | `127.0.0.1` | Address the callback server binds to |
| `MCP_OAUTH_REDIRECT_URI` | `http://localhost:$MCP_OAUTH_PORT/callback` | Redirect URI sent to the OAuth server |

Resolving once matters: the redirect URI has to be byte identical across client registration, the authorization request, and the token exchange. The bind address and the advertised address are separate because they differ behind port forwarding.

One thing that is not obvious from the issue: the registered redirect URI is now stored alongside the client ID. Dynamic Client Registration binds the URI to the client, and `authorize.ts` and `callback.ts` both validate the request against `client.redirectUris`. Since anyone hitting this bug has already run the CLI once on defaults, the cached client ID would be reused and the new redirect URI rejected, so the fix would not actually work. Clients are now re-registered when the URI changes. Clients registered before this change have no recorded URI and are assumed to use the previous default, so existing setups are untouched. I keyed on the stored URI rather than adding it to the config key because that key also holds the cached access token, and changing it would throw away valid tokens.

Overrides are validated: http or https only, no userinfo, ports 1-65535 (port 0 cannot work here since the browser needs the port before the server binds).

Non-loopback binds log a warning, since that exposes the callback server to the network.

## Testing

`pnpm run tsc` and `pnpm run lint` pass. `pnpm --filter @sentry/mcp-test-client run test` is 70 passed, including 20 new tests covering the resolver, the re-registration behavior, and that one redirect URI reaches registration, authorization, and token exchange. I confirmed the re-registration test fails against the old caching behavior.

I also ran a throwaway integration check (not committed, to keep real port binding out of CI) that started the callback server with `MCP_OAUTH_HOST=0.0.0.0` and an overridden redirect URI, fetched the callback over HTTP, and confirmed it resolved with the code and state.

`@sentry/mcp-agent-cli-test` has one failing test on my machine, but it fails the same way on a clean checkout of main (a Windows path issue), so it is unrelated to this change.

Per AGENTS.md, this was written with AI assistance and the commit carries the required `Co-Authored-By` attribution.

### DIFF
--- a/docs/testing/remote.md
+++ b/docs/testing/remote.md
@@ -195,6 +195,27 @@ pnpm -w run cli "who am I?"
 - Automatically refreshed when expired
 - To force re-auth: delete the token file
 
+**Running in a VM or container:**
+
+The callback server defaults to `127.0.0.1:8765`, which a browser on the host
+cannot reach. Three environment variables override it:
+
+| Variable | Default | Purpose |
+| --- | --- | --- |
+| `MCP_OAUTH_PORT` | `8765` | Port the callback server listens on |
+| `MCP_OAUTH_HOST` | `127.0.0.1` | Address the callback server binds to |
+| `MCP_OAUTH_REDIRECT_URI` | `http://localhost:$MCP_OAUTH_PORT/callback` | Redirect URI sent to the OAuth server |
+
+```bash
+# Bind on all interfaces and send the browser to the forwarded address
+MCP_OAUTH_HOST=0.0.0.0 \
+MCP_OAUTH_REDIRECT_URI=http://192.168.1.20:8765/callback \
+  pnpm -w run cli "who am I?"
+```
+
+Set `MCP_OAUTH_REDIRECT_URI` whenever the address the browser uses differs from
+the one the CLI binds to, such as behind port forwarding or a devcontainer.
+
 ### Direct Sentry Token Testing
 
 Remote `/mcp` also accepts explicit upstream Sentry API tokens with a separate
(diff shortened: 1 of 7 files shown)


dcramer:
Thanks!

PR #1243, opened 2026-08-12
fix(test-client): surface provider errors instead of silent empty response

### LINKED ISSUE #500: Errors in test client are not propagated to user, only show 'No response generated'
## Problem

When the MCP test client encounters errors (such as invalid API keys, rate limits, or API service errors), these errors are not properly propagated to the user. Instead, the client simply displays "(No response generated)" which doesn't help users understand what went wrong.

## Current Behavior

- Invalid OpenAI API key → Shows "(No response generated)"
- Rate limit errors → Shows "(No response generated)"  
- Invalid/unrecognized commands → Shows "(No response generated)"
- API service errors → Shows "(No response generated)"

## Expected Behavior

The client should display meaningful error messages to help users understand what went wrong:
- Invalid API key → "OpenAI API authentication failed. Please check your OPENAI_API_KEY environment variable."
- Rate limits → "OpenAI API rate limit exceeded. Please wait and try again."
- Service errors → "OpenAI API service error. The service may be temporarily unavailable."

## Technical Details

The issue appears to be related to how the Vercel AI SDK's `streamText` function handles errors. When an invalid API key is provided, the stream returns empty (no chunks) rather than throwing an error or including error events in the stream.

### Code Location
- File: `packages/mcp-test-client/src/agent.ts`
- Function: `runAgent`
- Lines: ~53-140

### Investigation Notes

1. The AI SDK's `streamText` function doesn't throw synchronously for API errors
2. The `textStream` iterator completes without yielding any chunks when there's an API error
3. Error information might be available in `result.fullStream`, `result.response`, or `result.errorPromise` but these are not currently being checked

## Reproduction Steps

1. Set an invalid OpenAI API key: `export OPENAI_API_KEY="sk-invalid-test-key"`
2. Run the test client: `pnpm run start:client`  
3. Enter any command (e.g., "test")
4. Observe that it shows "(No response generated)" instead of an authentication error

## Impact

Users cannot distinguish between:
- Configuration issues (wrong API key)
- Service issues (rate limits, outages)
- Invalid commands
- Actual empty responses

This makes debugging and troubleshooting very difficult.

zkasuran:
When the MCP test client hit a provider error (an invalid key, a rate limit or a 5xx), it printed `(No response generated)` and the real reason was lost. The Vercel AI SDK the repo pins (`ai@6`) does not throw from `streamText`: it ends `textStream` with zero chunks and reports the failure only through an `onError` callback the code never set. So the loop finished empty, `runAgent` never entered its `catch`, the Sentry span kept its OK status and nothing was captured. The failure was invisible to the user and to Sentry at once.

This sets `onError` to capture the failure, then rethrows it after the stream drains so it flows through the existing `catch`, which marks the span errored and rethrows to the CLI's `captureException`. `describeProviderError` maps 401/403 to auth, 429 to rate limit and 5xx to service, then unwraps `RetryError.lastError` for retryable failures. The legitimate empty-response path is untouched. Confirmed against the installed `ai@6.0.230`: `textStream` yields zero chunks without throwing and the error arrives via `onError` as an `APICallError`; the guess in the issue of `result.errorPromise` does not exist on this version.

Fixes #500

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally: the `@sentry/mcp-test-client` vitest suite (70 to 79 passing), the new tests failing before and passing after the fix, `tsc --noEmit`, Biome lint and format, ast-grep, plus a real Sentry send showing the error now reaches Sentry with the span marked errored.

### DIFF
--- a/packages/mcp-test-client/src/agent.test.ts
+++ b/packages/mcp-test-client/src/agent.test.ts
@@ -1,6 +1,23 @@
+import { APICallError, RetryError } from "ai";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { runAgent } from "./agent.js";
+import { describeProviderError, runAgent } from "./agent.js";
 import type { MCPConnection } from "./types.js";
+
+// Capture the status the Sentry span is given so tests can assert a provider
+// failure is reported as an error (code 2) rather than silently swallowed.
+const { spanStatuses } = vi.hoisted(() => ({
+  spanStatuses: [] as Array<{ code: number }>,
+}));
+
+vi.mock("@sentry/core", () => ({
+  startNewTrace: (cb: () => unknown) => cb(),
+  startSpan: (_opts: unknown, cb: (span: unknown) => unknown) =>
+    cb({
+      setStatus: (status: { code: number }) => {
+        spanStatuses.push(status);
+      },
+    }),
+}));
 
 function createStreamingResponse() {
   const encoder = new TextEncoder();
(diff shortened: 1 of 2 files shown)


dcramer:
testing some flows we have with junior here - will work on getting this merged

PR #1244, opened 2026-08-13
feat(tools): Add onboarding status updates

evanpurkhiser:
Add a catalog-only `onboarding_status_update` tool that reports explicit stage and run state to Sentry's agentic onboarding UI. The tool uses the existing region-aware API client with `org:read`, stays out of the default top-level tool context, and returns only whether subsequent progress updates should continue.

Stage failures remain retryable, while a failed or completed run is terminal. The optional `projectSlugs` and `issueIds` arrays are restricted to their owning stages and support incrementally reporting multiple validated projects or verification issues; Sentry accumulates and deduplicates those values across updates.

Calls are described as silent UI bookkeeping rather than analytics. The tool avoids echoing the run token, identifiers, or event notes, and generated definitions expose the same constrained request contract.

Depends on the agentic onboarding status API introduced in [getsentry/sentry#121914](https://github.com/getsentry/sentry/pull/121914) and its plural identifier response contract.

### DIFF
--- a/packages/mcp-core/src/tools/catalog/onboarding-status-update.ts
+++ b/packages/mcp-core/src/tools/catalog/onboarding-status-update.ts
@@ -1,0 +1,95 @@
+import { setTag } from "@sentry/core";
+import { z } from "zod";
+import {
+  AgenticOnboardingRunStatusUpdateSchema,
+  AgenticOnboardingStageSchema,
+  AgenticOnboardingStageStatusUpdateSchema,
+  AgenticOnboardingStatusUpdateSchema,
+} from "../../api-client/schema";
+import { apiServiceFromContext } from "../../internal/tool-helpers/api";
+import { defineTool } from "../../internal/tool-helpers/define";
+import { ParamOrganizationSlug, ParamRegionUrl } from "../../schema";
+import { ALL_SKILLS } from "../../skills";
+import type { ServerContext } from "../../types";
+
+export default defineTool({
+  name: "onboarding_status_update",
+  skills: ALL_SKILLS,
+  includeInSkillDefinitions: false,
+  requiredScopes: ["org:read"],
+  description: [
+    "Update the progress shown in Sentry's agentic onboarding UI.",
+    "",
+    "Use this tool only when the Sentry getting started skill provides a run token. Call it at the workflow boundaries described by that skill.",
+    "",
+    "<hints>",
+    "- Progress updates are operational UI state, not user analytics.",
+    "- Keep eventNote brief and limited to context useful in the progress UI.",
+    "- Do not include source code, repository paths, credentials, error output, or other sensitive data.",
+    "- status is required for every update.",
+    "- Set runStatus to completed or failed only when the entire onboarding run reaches that state.",
+    "- A failed status requires a brief eventNote explaining the failure.",
+    "</hints>",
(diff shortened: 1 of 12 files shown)


packages/mcp-core/src/server.test.ts
```diff
@@ -109,6 +109,7 @@ const DEFAULT_DIRECT_TOOL_NAMES = [
   "find_organizations",
   "find_projects",
   "get_sentry_resource",
+  "onboarding_status_update",
```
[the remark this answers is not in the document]
evanpurkhiser: This tool is idempotent for identical arguments. The backend builds the candidate state and returns the existing run unchanged when the candidate equals the current run; sequence increments only after that equality guard. The separate sequence values in these tests are independent fixtures, not repeated identical requests, and duplicate endpoint updates also suppress Conduit publishing. Keeping `idempotentHint: true`.

packages/mcp-core/src/tools/catalog/onboarding-status-update.ts
```diff
@@ -0,0 +1,94 @@
...
+import type { ServerContext } from "../../types";
+
+export default defineTool({
+  name: "onboarding_status_update",
+  skills: ALL_SKILLS,
+  requiredScopes: ["org:read"],
+  description: [
+    "Update the progress shown in Sentry's agentic onboarding UI.",
+    "",
+    "Use this tool only when an onboarding workflow provides a run token. Call it at the workflow boundaries described by the onboarding skill.",
```
evanpurkhiser: Probably should say the 'getting started' skill

packages/mcp-core/src/server.test.ts
```diff
@@ -3,7 +3,7 @@ import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
 import { McpServer as ModernMcpServer } from "@modelcontextprotocol/server";
 import { type Span, setUser, startSpan } from "@sentry/core";
 import { mswServer } from "@sentry/mcp-server-mocks";
-import { http, HttpResponse } from "msw";
+import { HttpResponse, http } from "msw";
```
evanpurkhiser: Why did this get reformatted

packages/mcp-core/src/skillDefinitions.json
evanpurkhiser, line 1: Why is it in here so many times?

packages/mcp-core/src/tools/catalog/onboarding-status-update.ts
evanpurkhiser: Updated the description to name the Sentry getting started skill explicitly.

packages/mcp-core/src/server.test.ts
evanpurkhiser: Reverted this unrelated import-order change.

packages/mcp-core/src/skillDefinitions.json
evanpurkhiser: The tool needs runtime availability under every granted MCP skill because the getting started skill lives outside the MCP server skill set. I separated runtime availability from generated skill documentation, so it remains catalog-discoverable but is no longer repeated in each generated skill bundle.

evanpurkhiser:
Alright looks good to go

PR #1266, opened 2026-08-25
fix(api): Support stage-specific onboarding metadata

evanpurkhiser:
Sentry onboarding runs now store project slugs and verification issue IDs in stage-specific `extra` objects. This updates the API schemas, outgoing status payloads, response parsing, and mocks to follow the contract introduced by getsentry/sentry#122076.

The `onboarding_status_update` tool now accepts a nested `update` object whose schema discriminates on `stage`. Callers must move `stage`, `status`, `runStatus`, and `eventNote` under `update`, and send `projectSlugs` or `issueIds` through that stage's `extra` object.

### DIFF
--- a/packages/mcp-core/src/api-client/schema.ts
+++ b/packages/mcp-core/src/api-client/schema.ts
@@ -2098,11 +2098,43 @@ export const AgenticOnboardingStageStatusUpdateSchema = z.enum([
   "failed",
 ]);
 
-export const AgenticOnboardingStageStateSchema = z.object({
-  stage: AgenticOnboardingStageSchema,
+const AgenticOnboardingCreateProjectExtraSchema = z
+  .object({
+    projectSlugs: z.array(z.string().trim().min(1)),
+  })
+  .strict();
+
+const AgenticOnboardingVerificationErrorExtraSchema = z
+  .object({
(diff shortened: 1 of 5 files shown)


packages/mcp-core/src/api-client/schema.ts
```diff
@@ -2098,10 +2098,28 @@ export const AgenticOnboardingStageStatusUpdateSchema = z.enum([
...
+export const AgenticOnboardingVerificationErrorExtraSchema = z
+  .object({
+    issueIds: z.array(z.string().trim().min(1)),
+  })
+  .strict();
+
+export const AgenticOnboardingStageExtraSchema = z.union([
+  AgenticOnboardingCreateProjectExtraSchema,
+  AgenticOnboardingVerificationErrorExtraSchema,
+]);
```
evanpurkhiser: Seems wrong, doesn't this need to actually discriminate on the `stage` key?

evanpurkhiser: Yes—the extra shape needs to be selected by stage. Updated both status updates and returned stage states to use discriminated unions on `stage`, with regression coverage for mismatched metadata.

*— Claude Code*

PR #1268, opened 2026-08-26
test(attachments): Cover binary MCP attachment responses

MathurAditya724:
Binary event attachments are now covered through both the SDK v2 catalog and Cloudflare Worker SSE paths, ensuring the returned MCP resource retains its MIME type and base64 body.

This is regression coverage only: the current source already preserves attachment bytes, while the reported live response requires deployment- and fixture-specific follow-up. Refs #1267.

### DIFF
--- a/packages/mcp-cloudflare/src/server/lib/mcp-handler.test.ts
+++ b/packages/mcp-cloudflare/src/server/lib/mcp-handler.test.ts
@@ -406,6 +406,51 @@ describe("MCP Handler", () => {
       });
     });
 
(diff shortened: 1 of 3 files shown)


MathurAditya724:
Closing this test-only PR: the deployed MCP server now succeeds for the reporter-provided attachment. `get_event_attachment` returns a valid non-error `image/png` result with 485,756 decoded bytes, matching the upstream attachment, and the listing no longer exposes SHA1. This branch is not needed to resolve the current production behavior; issue #1267 remains separate for triage.

PR #1277, opened 2026-08-31
chore: Upgrade to SDK v11 alpha

JPeer264:
This updates to the Sentry SDK v11 (alpha).

- `sendDefaultPii` has been removed in favor of `dataCollection` - it is even on by default
- `enableLogs` and `enableMetrics` were removed and are on by default

### DIFF
--- a/packages/mcp-cloudflare/src/client/instrument.ts
+++ b/packages/mcp-cloudflare/src/client/instrument.ts
@@ -4,7 +4,6 @@ import { resolveAttribution } from "./utils";
 
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
-  sendDefaultPii: true,
   tracesSampleRate: 1,
   beforeSend: sentryBeforeSend,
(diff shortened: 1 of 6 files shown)


packages/mcp-cloudflare/src/client/instrument.ts
```diff
@@ -4,7 +4,6 @@ import { resolveAttribution } from "./utils";
 
 Sentry.init({
   dsn: import.meta.env.VITE_SENTRY_DSN,
-  sendDefaultPii: true,
```
BYK, line 7: I thought we had a replacement setting for this now, where's that?

BYK (approved):

packages/mcp-cloudflare/src/client/instrument.ts
JPeer264: `dataCollection` is the replacement, but it has been changed to be on by default 

(https://blog.sentry.io/datacollection-control-panel/)

PR #1279, opened 2026-09-01
chore: Fix broken lock file after merge

JPeer264:
After merging #1277 it seems that the lock file was broken 🤷 

Edit: seems to be a combination of a not rebased #1277 on top of #1278


JPeer264:
Oh. Seems like no approval is needed that it's getting merged.

PR #1280, opened 2026-09-01
fix: Enable gen_ai spans with vite plugin

JPeer264:
With v11 we move over to orchestrion. Since the MCP is using `ai@6` it needs build time instrumentations to actually receive spans for libraries that do not support tracing channels OOTB.

Currently the AI dashboards are not receiving data (as there is also none sent). This plugin adds these

### DIFF
--- a/packages/mcp-cloudflare/src/server/sentry.config.ts
+++ b/packages/mcp-cloudflare/src/server/sentry.config.ts
@@ -21,7 +21,7 @@ export default function getSentryConfig(env: Env): CloudflareOptions {
     environment:
       env.SENTRY_ENVIRONMENT ??
       (process.env.NODE_ENV !== "production" ? "development" : "production"),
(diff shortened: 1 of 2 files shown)


packages/mcp-cloudflare/src/server/sentry.config.ts
```diff
@@ -21,7 +21,7 @@ export default function getSentryConfig(env: Env): CloudflareOptions {
     environment:
       env.SENTRY_ENVIRONMENT ??
       (process.env.NODE_ENV !== "production" ? "development" : "production"),
-    integrations: [Sentry.zodErrorsIntegration(), Sentry.vercelAIIntegration()],
```
JPeer264, line 24: note: The integrations are automatically added when the package is detected

JPeer264:
Going to ninja merge since the `gen_ai` spans are missing

PR #1281, opened 2026-09-02
fix(attachments): prevent worker OOM on large event attachments

betegon:
## Summary

`get_event_attachment` crashes the Cloudflare Worker with `TypeError: Memory limit exceeded before EOF` when downloading large attachments. This is live and ongoing — Sentry issue [MCP-SERVER-F5W](https://sentry.sentry.io/issues/7284551639/): **161 events, 10 users**, first seen 2026-02-22, still firing (last 2026-09-01) across orgs and clients (`codex`, `cursor`).

Root cause: the download buffered the file **~4–5× in memory at once** — `downloadResponse.blob()` → `blob.arrayBuffer()` (a copy) → base64 (`+33%`) → JSON-serialization of the base64 — blowing the 128 MB isolate on the tens-of-MB tail. Real download sizes (90d span telemetry): p50 ≈ 90 KB, p99 ≈ 16 MB, **max ≈ 38 MB**.

Refs #1267.

## Changes

- **Halve peak memory:** read the body once via `arrayBuffer()` and base64-encode from a `Buffer` view (`bytesToBase64`, no intermediate `Blob`/copy). Peak drops from ~4–5× to ~2× the payload.
- **Inline size cap (30 MB):** gated on the already-fetched metadata `size`, so an oversized attachment never issues the doomed download. 30 MB → ~80 MB peak, safely under the isolate; covers p99 and all observed traffic inline.
- **Usable oversized fallback:** the download URL is an authenticated API endpoint (bare fetch → 401), so a plain link is a dead end for agents. For **objectstore** attachments (the modern default) we resolve the **presigned URL** from the download endpoint's `302 Location` — a self-authenticating, time-limited link that downloads with **no credentials**. Legacy filestore attachments (no redirect) fall back to authenticated-download instructions (`sentry api …`).
- **Empty/truncated guard:** a zero-byte body when metadata reports a non-empty file is now surfaced instead of emitting empty content.

## Verification

Confirmed end-to-end against a real attachment (created in a test org):
1. Authenticated download → `302` to an objectstore presigned URL.
2. That presigned URL fetched with **no auth header** → `200` + the correct bytes.

Also confirmed the download endpoint is auth-gated (401/403 unauthenticated) and validated the redirect vs. streaming behavior against `getsentry/sentry` source.

## Testing

`pnpm tsc`, `biome lint`/`format`, `ast-grep`, and the full `mcp-core` suite pass. New coverage: text (`.log`) attachments, oversized objectstore (→ presigned URL), oversized legacy (→ auth instructions), and empty body.

### DIFF
--- a/packages/mcp-core/src/api-client/client.ts
+++ b/packages/mcp-core/src/api-client/client.ts
@@ -256,6 +256,14 @@ type RequestOptions = {
   allowStatuses?: number[];
 };
 
+/**
+ * Default cap for returning attachment bytes inline over the MCP transport.
+ * A 30 MB attachment takes ~80 MB in memory once base64-encoded and serialized,
+ * which fits the Cloudflare Worker's 128 MB isolate with room for the rest of the
+ * request; above it, getEventAttachment returns a reference instead of the bytes.
+ */
+export const DEFAULT_MAX_INLINE_ATTACHMENT_BYTES = 30 * 1024 * 1024;
+
 export type TraceItemType = "spans" | "logs" | "tracemetrics";
 
 type ClientKeyRateLimit = {
@@ -3839,16 +3847,16 @@ export class SentryApiService {
       projectSlug,
       eventId,
       attachmentId,
-      maxInlineBytes,
+      maxInlineBytes = DEFAULT_MAX_INLINE_ATTACHMENT_BYTES,
     }: {
       organizationSlug: string;
       projectSlug: string;
       eventId: string;
       attachmentId: string;
       /**
-       * When set, attachments larger than this (by metadata `size`) are not
-       * downloaded; the method returns `bytes: null` so the caller can hand back
-       * a download reference instead.
+       * Attachments larger than this (by metadata `size`) skip download —
+       * `getEventAttachment` returns `bytes: null`. Defaults to
+       * {@link DEFAULT_MAX_INLINE_ATTACHMENT_BYTES}.
        */
       maxInlineBytes?: number;
     },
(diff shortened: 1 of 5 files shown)


packages/mcp-core/src/api-client/client.ts
```diff
@@ -3887,15 +3920,47 @@ export class SentryApiService {
...
+   * discarded either way — we only need the URL.
+   */
+  private async resolveDirectDownloadUrl(
+    downloadPath: string,
+    opts?: RequestOptions,
+  ): Promise<string | null> {
+    try {
+      const response = await this.request(
+        downloadPath,
+        { method: "GET" },
```
BYK, line 3959: Why not a HEAD request at this point instead?

packages/mcp-core/src/api-client/client.ts
```diff
@@ -3839,18 +3839,32 @@ export class SentryApiService {
...
       organizationSlug: string;
       projectSlug: string;
       eventId: string;
       attachmentId: string;
+      /**
+       * When set, attachments larger than this (by metadata `size`) are not
+       * downloaded; the method returns `bytes: null` so the caller can hand back
+       * a download reference instead.
+       */
+      maxInlineBytes?: number;
```
BYK, line 3861: Shouldn't we have a default value for this?

BYK (approved):

packages/mcp-core/src/api-client/client.ts
betegon: yes. added `DEFAULT_MAX_INLINE_ATTACHMENT_BYTES` (30MB) as the default

packages/mcp-core/src/api-client/client.ts
betegon: tested it but it gives us a 403 for this endpoint while the GET gives the 302→presigned. had a look and HEAD is in the allow header but the objecstore presign path only works for GET

PR #1284, opened 2026-09-03
feat(formatter): return structuredContent from get_issue_details

shayna-ch:
Migrates get_issue_details to the structured result path. The tool declares an outputSchema and returns structuredResult(payload) once the event carries a json body from sentry's shared formatter, so the server generates the compatibility text rather than the tool hand-assembling markdown.

The event endpoint now requests ?llmFormat=json. formatIssueOutput's shared formatter branch is guarded on format === "markdown" so a json body is never pasted into prose; the autofix endpoint keeps requesting markdown until analyze_issue_with_seer migrates too.

event.body is an open record on purpose. It is whatever the shared formatter emits, and its sections are decided there: enumerating them here would make this schema a second declaration of that contract, needing a bump every time a section lands on the sentry side.

Markdown is still returned when there is no json body, which is what an org that is not on the rollout gets. structuredContent has to carry the whole answer, and without the event it would not, so this follows the "markdown only until there is a real structured payload" half of the packing policy rather than shipping a partial payload. Not found on an event id lookup now throws UserInputError like the tool's other input failures, since prose is not a structured result.

Seer state is summarized to status plus the root cause and solution one liners rather than embedded whole: an AutofixRunState carries every step and would dwarf the rest of the payload.


shayna-ch:
bugbot run

shayna-ch:
bugbot run

shayna-ch:
bugbot run

shayna-ch:
bugbot run

gricha (approved):
lets test it before opening to public!

PR #1286, opened 2026-09-04
feat(tools): Add get_sentry_mcp_info catalog tool

### LINKED ISSUE #1148: A tool that returns the current version of the MCP server
I've run into an issue recently where I have a Skill that I created and is used by many of my teammates stopped working as expected. The reason was that some of the tool names here changed and some tools split into multiple tools.

If I had a way to determine the current version of the MCP that is being accessed I can alert the user that they are using an MCP version out of sync with the Skill's expectations. And possibly stop the Skill

A tool in the MCP that returned its version would allow me to add logic to my Skill to make sure things are all in sync and happy with each other.

jane-jhu:
## Summary

- Add a catalog-only `get_sentry_mcp_info` tool
- Return the running Sentry MCP server version using the existing `LIB_VERSION`
- Make the tool available to all Skills without requiring Sentry API scopes
- Regenerate tool and Skill definitions
- Add unit and server-level coverage for catalog discovery and execution

This allows Skills and integrations to check the running Sentry MCP version and verify compatibility.
Fixes #1148

## Test plan

- `pnpm run tsc`
- `pnpm run lint`
- `pnpm run test`
- Verified that `search_sentry_tools` finds `get_sentry_mcp_info` for `server version`
- Verified that `execute_sentry_tool` returns:

```json
{
  "version": "0.39.0"
}

### DIFF
--- a/packages/mcp-core/src/server.test.ts
+++ b/packages/mcp-core/src/server.test.ts
@@ -17,6 +17,7 @@ import {
 import { createExecuteTool } from "./tools/special/execute-tool";
 import type { ToolConfig } from "./tools/types";
 import type { ServerContext } from "./types";
+import { LIB_VERSION } from "./version";
 
 // Mock the Sentry core module
 vi.mock("@sentry/core", () => ({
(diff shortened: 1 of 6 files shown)


dcramer (approved):

dcramer:
Thanks!

PR #1296, opened 2026-09-08
fix(issue details): cap related replays in the structured payload

shayna-ch:
A real issue returned 51 related replay ids. The markdown output shows a Related Replay Count plus the first MAX_DISPLAY_REPLAYS and an "and N more" line, so the payload was the one place every id landed. Caps at the same 5 and carries relatedCount so nothing is hidden by the cap.

Found by calling the tool against a live issue rather than a fixture, which is also why the count was missing: no test had more than a couple of replays.


shayna-ch:
bugbot run

shayna-ch:
bugbot run

gricha (approved):

PR #1315, opened 2026-09-17
test(evals): add representative natural-language search coverage

skaasten:
## Summary

Add compact natural-language search eval coverage across errors, logs, spans, metrics, and issues.

### Key Changes

- Add one representative scenario for each supported search dataset
- Validate translated datasets, filters, fields, sorting, and time ranges
- Assert attribute discovery for the metrics scenario

### Breaking Changes

- None


betegon (approved):
thanks for adding these! 

we had a bunch of issues over time series in the past (until we merged #1305 a couple of weeks ago), so it's good to add more coverage.

